Issue #44, opened 2022-09-30
Handle `pg_dump` / `pg_restore` workflow when TLE extensions are installed


jkatz:
Currently, a `pg_dump` file will not put out TLE artifacts in the correct install order. For example, if we have a TLE extensions called "tle_test", `pg_dump` will generate something like this:

```
CREATE EXTENSION pg_tle;
CREATE EXTESNION tle_test;

CREATE FUNCTION pgtle."tle_test.control()" ...
```

This will fail, as the "control file" is not available yet in the DB.

One option is to provide a `pg_tle_dump` wrapper script around `pg_dump` to put the TLE artifacts in the correct order.

formalLogicGirl:
In general, pg_dump creates extension objects before function/procedure objects, which is why we have this behavior.

I have a viable path for fixing this. I did a prototype implementation to confirm that adding an entry to pg_depend for the .control file in tleCreateExtension() makes pg_dump do the correct ordering, with the creation of the extension file after the function it depends on. pg_dump output from my implementation looks like the following, note that function test123.control will be created before extension test123.

` --
-- PostgreSQL database dump
---- Dumped from database version 14.6
-- Dumped by pg_dump version 14.6SET statement_timeout = 0;
SET lock_timeout = 0;
SET idle_in_transaction_session_timeout = 0;
SET client_encoding = 'UTF8';
SET standard_conforming_strings = on;
SELECT pg_catalog.set_config('search_path', '', false);
SET check_function_bodies = false;
SET xmloption = content;
SET client_min_messages = warning;
SET row_security = off;--
-- Name: pgtle; Type: SCHEMA; Schema: -; Owner: swmiller
--CREATE SCHEMA pgtle;
ALTER SCHEMA pgtle OWNER TO swmiller;--
-- Name: pg_tle; Type: EXTENSION; Schema: -; Owner: -
--CREATE EXTENSION IF NOT EXISTS pg_tle WITH SCHEMA pgtle;
--
-- Name: EXTENSION pg_tle; Type: COMMENT; Schema: -; Owner:
--COMMENT ON EXTENSION pg_tle IS 'Trusted Language Extensions for PostgreSQL';
--
-- Name: test123.control(); Type: FUNCTION; Schema: pgtle; Owner: swmiller
--
CREATE FUNCTION pgtle."test123.control"() RETURNS text
    LANGUAGE sql
    AS $_X$SELECT $_pgtle_i_$default_version = '1.1'
[...]

PR #151, opened 2022-12-08
Record dependency of a TLE on its control and sql functions

formalLogicGirl:
Issue #, if available:
#44 

Description of changes:

Record dependency of an extension installed by pg_tle on its control function and sql function, when the extension is created. This is in for pg_dump to correctly create the control and sql functions before creating the extension. Otherwise, the default order of object creation by pg_dump is to create functions after extensions, so that restoring from a dump will fail to create TLE-installed extensions.

### DIFF
--- a/src/tleextension.c
+++ b/src/tleextension.c
@@ -22,6 +22,7 @@
  */
 #include "postgres.h"
 
+#include <assert.h>
 #include <dirent.h>
 #include <limits.h>
 #include <sys/file.h>
@@ -52,6 +53,7 @@
 #include "catalog/pg_depend.h"
 #include "catalog/pg_extension.h"
 #include "catalog/pg_namespace.h"
+#include "catalog/pg_proc.h"
 #include "catalog/pg_type.h"
 #include "commands/alter.h"
 #include "commands/comment.h"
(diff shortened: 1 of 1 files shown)


jkatz:
@formalLogicGirl Can you add a test? I believe this would be through the TAP tests.

formalLogicGirl:
> @formalLogicGirl Can you add a test? I believe this would be through the TAP tests.

Yes. I'll add tap tests as a separate PR, after I get the full workflow of restoring from dump well tested. Don't want to hold up too much stuff into one. I'll have multiple PRs for this issue.

formalLogicGirl:
@jconway Is there a better way to get function OID? I'm using SPI. Something similar to SearchSysCache() thats exposed for extension code to call?

formalLogicGirl:
> @jconway Is there a better way to get function OID? I'm using SPI. Something similar to SearchSysCache() thats exposed for extension code to call?

will use LookupFuncName() and update this PR. Thanks.

formalLogicGirl:
As an alternative to SPI_execute to look up function oids, I tried the following

```
        char *qualifiedCtlNameStr = psprintf("%s.%s.control", PG_TLE_NSPNAME, extname);
        ctlFuncId = LookupFuncName(stringToQualifiedNameList(qualifiedCtlNameStr), 0, NULL, false /* missing_ok */);
```

However, that doesn't work because pgtle.extname.control is parsed as database name pgtle, schema name extname, function name control. Having . in the function name, which is not disallowed by postgres, causes ambiguity in parsing a schema-qualified name in the postgres functions used here.

jkatz (approved):

formalLogicGirl:
This fixes one of the ordering problems in pg_dump generated TOC. There are some more problems for the pg_dump/pg_restore workflow to work, and I'm continuing to investigate and work on them in #154 and get #152 to work. This PR contributes one part of #44

formalLogicGirl:
Closed by the following

https://github.com/aws/pg_tle/pull/151 - pg_dump working with pg_tle
https://github.com/aws/pg_tle/pull/155 - pg_restore working with pg_tle
https://github.com/aws/pg_tle/pull/157 - test
https://github.com/aws/pg_tle/pull/156 - doc update

PR #162, opened 2022-12-23
Changes for version 16 and above acl checks were refactored in the server. 

davecramer:
Issue #, if available:

Description of changes:

### DIFF
--- a/include/compatibility.h
+++ b/include/compatibility.h
@@ -172,4 +172,15 @@ _PU_HOOK;
 #endif
 #endif
 
+#if (PG_VERSION_NUM < 160000)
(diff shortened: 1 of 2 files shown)


jkatz:
@davecramer would it be possible to put this change in compatibility.h and use that to incorporate the updated function definition?

davecramer:
> @davecramer would it be possible to put this change in compatibility.h and use that to incorporate the updated function definition?

Of course. However there are already a number of places where we check for version numbers. I was copying the existing code. I could see using macros though.

jconway:
Agree with Jonathan. At one point I think all of the version dependencies were centralized in compatibility.h. IMHO, to the extent that has eroded over time, it should be fixed rather than continued.

davecramer:
fair enough

jkatz (approved):

PR #177, opened 2023-02-21
Feature split

jim-mlodgenski:
Description of changes: Split out some of the code for managing features that are common into its own file. 

### DIFF
--- a/include/feature.h
+++ b/include/feature.h
@@ -56,6 +56,6 @@ static const struct config_enum_entry feature_mode_options[] = {
 
 #define FEATURE_TABLE "feature_info"
 
-void feature_proc(List **proc_names, const char *featurename);
+List * feature_proc(const char *featurename);
 
 #endif							/* FEATURE_H */
--- a/include/feature.h
+++ b/include/feature.h
@@ -1,0 +1,61 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * feature.h
+ */
+
+#ifndef FEATURE_H
+#define FEATURE_H
+
+/*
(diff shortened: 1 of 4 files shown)


src/feature.c
```diff
@@ -0,0 +1,155 @@
...
+		query = psprintf("SELECT schema_name, proname FROM %s.%s WHERE feature OPERATOR(pg_catalog.=) $1::%s.pg_tle_features ORDER BY proname",
+			 quote_identifier(PG_TLE_NSPNAME), quote_identifier(FEATURE_TABLE), quote_identifier(PG_TLE_NSPNAME));
+		featargs[0] = CStringGetTextDatum(featurename);
+
+		ret = SPI_execute_with_args(query, 1, featargtypes, featargs, NULL, true, 0);
+
+
+		if (ret != SPI_OK_SELECT)
+			ereport(ERROR,
+					errmsg("Unable to query \"%s.feature_info\"", PG_TLE_NSPNAME));
```
jkatz, line 76:
```suggestion
					errmsg("Unable to query \"%s.%s\"", PG_TLE_NSPNAME, FEATURE_TABLE));
```

include/feature.h
```diff
@@ -0,0 +1,61 @@
...
+ * also attempt to provide some flexibility for use-cases that may be database specific.
+ *
+ */
+
+typedef enum enable_feature_mode
+{
+	FEATURE_ON,					/* Feature is enabled at the database level if
+								 * the entry exists */
+	FEATURE_OFF,				/* Feature is not enabled at the cluster level */
+	FEATURE_REQUIRE				/* Feature is enabled in all databases, errors
```
jkatz, line 46: minor comment: Should we prefix with `PGTLE_`? I don't think it's likely to conflict with anything in PG, but just in case...

I know we have this as `FEATURE_` currently, so OK either way.

jkatz (commented):
Generally looks good. This should make it trivial to add new features.

Left a few minor comments.

include/feature.h
jim-mlodgenski: Maybe, but we can do that as a separate PR

formalLogicGirl:
Looks good to me!

formalLogicGirl (approved):

include/feature.h
jkatz: 👍

src/feature.c
```diff
@@ -0,0 +1,155 @@
...
+		PG_RE_THROW();
+	}
+	PG_END_TRY();
+
+	return procs;
+}
+
+/*  Check for semi-colon to prevent SPI_exec from running multiple queries accidentally */
+static bool
+check_valid_name(char *val, const char *featurename)
```
jkatz, line 131: I noticed this change from `void` to `bool` -- the function either errors or returns, so unsure why we need to return a `bool`. It's `static` so not super vested either way, but I just noticed the change.

jkatz (approved):

PR #178, opened 2023-03-01
Docs: supporting MacOS

kiwicopple:
I tried following the docs and immediately ran into an error `Long arguments are not supported on MACOS`

This change is a suggestion to add better MacOS support. If you agree with the approach, I can do the same for the other commands.

### DIFF
--- a/tools/pg_tle_manage_local_extension/README.md
+++ b/tools/pg_tle_manage_local_extension/README.md
@@ -75,7 +75,8 @@ OPTIONS:
 
     -p, --extpath <Extension Path>
           A required parameter for install action.
-          Local path of the extension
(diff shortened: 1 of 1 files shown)


jkatz:
@kiwicopple Ah nice! That's why I had trouble running that on my Mac 😄 

+1 for this approach.

kiwicopple:
I've been travelling for the past few weeks @jkatz - I won't get a chance to finish the PR this month. Feel free to close or merge

jkatz:
@kiwicopple Thanks. I'm leaning towards merging but need to do a quick test on my Mac :smile:

jkatz (approved):

PR #192, opened 2023-04-14
Better error handling when an extension doesn't exist

### LINKED ISSUE #187: Error message creating non-existent extension with pg_tle installed
## Current Behavior
When pg_tle is installed abut not created and a user attempts to create an extension that does not exist e.g.

```sql
create extension not_available;
```
It returns a misleading error message:
```
ERROR:  schema "pgtle" does not exist
```
rather than the default
```
ERROR:  extension "not_available" is not available
```
That error message goes away if pg_tle is created, or created and dropped i.e.
```
create extension pg_tle;
drop extension pg_tle;
create extension not_available;'
-- ERROR:  extension "not_available" is not available
```

## Expected Behavior

Creating a non-existent extension on a fresh db with pg_tle installed but not created does not impact error messages.

## Local Reproduction Case

Start a fresh instance of postgres with pgt_tle installed, but not created:
```sh
docker run -it \
  -e POSTGRES_PASSWORD=password \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_DB=db \
  -p 4440:5432 \
  -d supabase/postgres:15.1.0.65
```

Connect to the database and try to create an extension that doesn't exist
```sh
PGPASSWORD=password PGHOST=0.0.0.0 psql \
  -p 4440 \
  -U postgres \
  -c 'create extension does_not_exist;'
```
Returns `ERROR:  schema "pgtle" does not exist`

jkatz:
If the shared preload library for "pg_tle" is loaded but the "pg_tle" extension is not activated and a user tries to install an extension that does not exist, pg_tle would give an error of the form "cannot find pgtle schema". However, while true, this is not a helpful error message.

The issue was due to the ProcessUtility hook firing even though the "pg_tle" extension itself was not active. To avoid this, we add a check to the PU hook to only run its custom logic when we know that "pg_tle" is installed.

fixes #187


jkatz:
Some notes to consider:

* Move the load checks to a function to better encapsulate.
* Consider moving the standard PU execution portion to its own call (though note the compatibility layer in place)
* Add tests 😄

PR #197, opened 2023-05-22
Support building pg_tle with Postgres 16devel sources

gurjeet:
Before this patch, compiling pg_tle against master branch (Postgres
16devel) leads to compiler errors, complaining about absence of function
get_extension_schema().

This commit improves the preprocessor macro's test condition to better
target the Postgres commits that don't contain the relevant change in
Postgres commit history.

### DIFF
--- a/src/tleextension.c
+++ b/src/tleextension.c
@@ -381,14 +381,48 @@ funcstat(char *procedureName)
 	return found;
 }
 
-#if (PG_VERSION_NUM < 160000)
+#if (PG_VERSION_NUM < 160000) || (CATALOG_VERSION_NO < 202303081)
 /*
  * get_extension_schema - given an extension OID, fetch its extnamespace
  *
  * Returns InvalidOid if no such extension.
  *
- * Note: e20b1ea157 makes this an external function, so we do not need
- * to define this for newer version of PostgreSQL.
+ * Note: This code is borrowed from Postgres sources. Postgres' commit
+ * e20b1ea157 makes this an external function, so we do not need to define this
+ * function for Postgres versions 16 and later.
+ *
+ * But while Postgres version 16 is in development, we need a way to build this
+ * extension against the pre-release versions of Postgres 16. While in
+ * development, the value of PG_VERSION_NUM is not granular enough to tell us
+ * if the commit we're interested in is present in the Postgres development
+ * version we're working with.
+ *
+ * Ideally, when building against Postgres commit e20b1ea157, or any descendant
+ * commits, we should not define this function. But we do not have the ability
+ * to target a range of commits in the preprocessor directives. So we use the
+ * Catalog version identifier as a proxy to get as close to that commit, as
+ * possible. Postgres commit 30a53b7929 is the first commit after e20b1ea157
+ * that changes the Catalog version number, and sets it to 202303081. So we use
+ * that Catalog version number in the preprocessor condition, to demarcate
+ * which range of commits in v16 development version should this function be
+ * hidden from, and for which ones we should define it.
+ *
+ * There are still 14 commits in the range [e20b1ea157, 30a53b7929) that this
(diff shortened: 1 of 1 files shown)


jconway (commented):
The version number check is the normal way to do this. Why add the catalogue check?
 -1

jkatz:
Agree with @jconway -- as long as this continues to work against HEAD I don't see any reason to change this. If something changes in HEAD that affects compatibility, we should fix it.

gurjeet:
The core concern is that if an extension supports PG version N, then it should support version N+1 as well, even while N+1 is in development mode. pg_tle supports Postgres version 15, but it cannot be built with many commits that are in 16devel.

A full reasoning is provided in the comments in the file. See [src/tleextension.c#L394](https://github.com/aws/pg_tle/blob/2522892b5077212c443dffaa5c57063fdf9e76ce/src/tleextension.c#L394) in the patch.

In short, though, pg_tle supports Postgres 15, but there are many commits in Postgres master branch (Postgres 16devel) that pg_tle fails to build against. They all error out with the following error.

    src/tleextension.c:3622:16: error: implicit declaration of function 'get_extension_schema' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                            reqschema = get_extension_schema(reqext);

The check `PG_VERSION_NUM < 16` evaluates to false for any commit after 16devel was stamped.

> d31d30973a1 2022-06-30 Stamp HEAD as 16devel.

This patch is an attempt to build pg_tle with as many commits of the master (16devel) branch, as possible.

Ideally, we'd like the ability to check for a specific commit/change number in preprocessor directives, to precisely identify a point in development history that includes the change we're interested in. But I also get the point that an out-of-tree extension should not bother itself with devel branches of Postgres, because devel, Beta, and RC versions will never be encountered in production.

So I guess the message is that extension developers should wait for at least Beta releases of Postgres to begin coding against the newer version, and not try to build against devel versions.

To that end, I tried building pg_tle against Beta versions of Postgres 15, and it failed to build against 15 Beta1 and 15 Beta2. This is documented in PR #198.

But, to be honest, as an extension developer I would like to build and test my extension with every commit, to ensure that there are no breaking changes in Postgres. For example, Postgres commit a2095f7fb5 fixed a bug that affected Index Advisor extension, and I would have liked to know that breakage as soon as it was introduced in the master/devel branch of Postgres 9.0.

I intend to raise this concern in -hackers, and see if there is a better technique we can adopt to let out-of-tree extensions build with in-development versions of Postgres. That will require being able to identify commits/ranges with a monotonically increasing integer value, similar to `CATALOG_VERSION_NO`.

For another example why we might need the ability to identify certain commits, my Postgres build script fails to run configure unchanged, ever since a the recent introduction of icu-support-by-default change. That is easier to handle since mine is a Bash function that can acccept `--without-icu` flag, and it has access to all the Postgres source code and Git history. But the extensions do not have this luxury, and hence have to stop developing and building against master/HEAD until the development branch is in promoted to Beta.

gurjeet:
Having said the above, feel free to close this issue, because I feel it's a larger issue that needs to be addressed by Postgres, and not the extension developers.

jkatz:
> To that end, I tried building pg_tle against Beta versions of Postgres 15, and it failed to build against 15 Beta1 and 15 Beta2.

Given PG15 is GA, there's no reason why this has to work against old beta versions. Even during the beta period, as soon as a new beta comes out, all the previous betas are mooted. Ensuring an extension works against previous betas is not a requirement.

If the main issue is we need to fix compilation against an unreleased version of PostgreSQL, that's fine -- and in fact that's relatively normal. While the community tries to keep the ABI stable after beta 1, things can (and do) change to help with the stability of the release. Once we GA, typically the ABI stays stable, though there have been changes known to occurs. That's in part why we keep a CI/CD with `pg_tle` to help ensure we continue building.

Closing this PR.

PR #198, opened 2023-05-22
Make a comment more technically correct

gurjeet:
The macro definition set_config_option_ext(), protected by the
preprocessor check 'PG_VERSION_NUM < 150000', allows us to build pg_tle
with Postgres versions 15 Beta3 and above, but it does not help when
building with Postgres 15beta2 or earlier.

### DIFF
--- a/include/compatibility.h
+++ b/include/compatibility.h
@@ -97,9 +97,9 @@
 #endif
 
 /*
- * PostgreSQL 15 introduces the ability to assign permissions to adjust server
- * variables. This adds the call for the new function in previous PostgreSQL
- * versions.
+ * PostgreSQL 15 Beta3 (commit a2944d8724) introduces the ability to assign
+ * permissions to adjust server variables. This adds the call for the new
+ * function in PostgreSQL 14 and earlier.
  */
 #if PG_VERSION_NUM < 150000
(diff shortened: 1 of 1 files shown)


jkatz:
I'm OK to specify the commit in the comment, but the "Beta 3" part is not needed. Each beta release is mooted as soon as the new beta comes out, so it doesn't matter which beta release it's in. What matters is the major.

gurjeet:
Comment updated and removed the commit-id, as requested.

jkatz:
That's the opposite of what I said:

> I'm OK to specify the commit in the comment, but the "Beta 3" part is not needed.

gurjeet:
Sorry, I read you message to mean the exact opposite of what you said. Must've
been distracted at the moment.

Give your clarification of how to treat Postgres Betas, and reading the
[original patch][] of this PR, I don't feel it adds anything useful for the
reader.

The patch was initially proposed to accompany PR #197, where I had discovered a
problem for extension developers, and this change in comment would have served
as an example of that discovered problem. Since PR #197 was closed without
merging (because that concern is better addressed by Postgres community), I am
closing this PR as well.

[original patch]:
https://github.com/aws/pg_tle/pull/198/commits/1589140ee99c2e2b82d7eded4398edf804f6ccf0

PR #203, opened 2023-06-08
Removed dead/unreachable return statements

gurjeet:
Code after ereport(ERROR) and elog(ERROR) calls is unreachable, so
remove such code.


gurjeet:
As mentioned in my closing comment of PR #204, when making this proposal I did
not consider the fact that pg_tle uses code heavily borrowed from Postgres.
Hence closing this PR for the same reason as PR #204.

Since this PR has actual code changes (vs. PR #204's formatting changes),
please feel free to reopen it, if you think these changes don't affect the
complexity of future merges from Postgres sources.

PR #204, opened 2023-06-08
Code formatting to improve readability

gurjeet:
Add or remove blank lines, add curly braces, etc., to improve
readability.

This is not a full sweep of the code base, just for the parts that I was
inspecting to look for dead code.

### DIFF
--- a/src/tleextension.c
+++ b/src/tleextension.c
@@ -490,6 +490,7 @@ check_valid_extension_name(const char *extensionname)
 				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
 				 errmsg("invalid extension name: \"%s\"", extensionname),
 				 errdetail("Extension names must only contain alphanumeric characters or valid separators.")));
+
 		idx++;
 	}
 }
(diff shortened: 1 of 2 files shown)


src/tleextension.c
```diff
@@ -2926,9 +2927,11 @@ pg_extension_config_dump(PG_FUNCTION_ARGS)
 			arrayLength < 0 ||
 			ARR_HASNULL(a) ||
 			ARR_ELEMTYPE(a) != OIDOID)
+        {
 			elog(ERROR, "extconfig is not a 1-D Oid array");
-		arrayData = (Oid *) ARR_DATA_PTR(a);
+        }
```
jkatz, line 2932: This is taken directly from upstream and was a deliberate design decision -- we wanted to keep the style close to that to simplify future merging:

https://github.com/postgres/postgres/blob/53ea2b7ad050ce4ad95c89bb55197209b65886a1/src/backend/commands/extension.c#L2504

src/tleextension.c
```diff
@@ -3064,9 +3070,11 @@ extension_config_remove(Oid extensionoid, Oid tableoid)
...
```
jkatz, line 3075: Similarly, this is taken directly from upstream to simplify any future merging efforts.

gurjeet:
> This is taken directly from upstream and was a deliberate design decision -- we wanted to keep the style close to that to simplify future merging:

Even though I mentioned it in another PR's comment, I did not consider this
fact when proposing the patch. Given that a lot of code was borrowed from
Postgres, I believe it's best if we don't fix formatting, etc. to keep future
merges simple.

Please disregard this proposal. Closing the PR.

PR #205, opened 2023-06-08
Implement pgtle.create_shell_type function to support shell type creation in pg_tle

lyupan:
Description of changes:
Implement pgtle.create_shell_type/create_shell_type_if_not_exists function:
1. They require two arguments: namespace and type name and can only be executed by pgtle_admin and superuser;
2. Similar to https://github.com/aws/pg_tle/blob/main/docs/03_managing_extensions.md#pgtleregister_feature_if_not_existsproc-regproc-feature-pg_tle_features, create_shell_type_if_not_exists returns false if the type already exists while create_shell_type fails;
3. Add a new pg_tle version 1.0.5 for the new datatype functions.

This is the first commit to support custom base types in pg_tle.

### DIFF
--- a/pg_tle--1.0.1--1.0.4.sql
+++ b/pg_tle--1.0.1--1.0.4.sql
@@ -107,26 +107,3 @@ GRANT EXECUTE ON FUNCTION pgtle.install_extension_version_sql
   version text,
   ext text
 ) TO pgtle_admin;
-
-CREATE FUNCTION pgtle.create_shell_type
-(
-  p_typnamespace regnamespace,
-  p_typname name
-)
-RETURNS boolean
-SET search_path TO 'pgtle'
-STRICT
-AS 'MODULE_PATHNAME', 'pg_tle_create_shell_type'
-LANGUAGE C;
-
-REVOKE EXECUTE ON FUNCTION pgtle.create_shell_type
-(
-  p_typnamespace regnamespace,
-  p_typname name
-) FROM PUBLIC;
-
-GRANT EXECUTE ON FUNCTION pgtle.create_shell_type
-(
-  p_typnamespace regnamespace,
-  p_typname name
-) TO pgtle_admin;
--- a/pg_tle--1.0.1--1.0.4.sql
+++ b/pg_tle--1.0.1--1.0.4.sql
@@ -107,3 +107,26 @@ GRANT EXECUTE ON FUNCTION pgtle.install_extension_version_sql
   version text,
   ext text
 ) TO pgtle_admin;
+
+CREATE FUNCTION pgtle.create_shell_type
+(
+  p_typnamespace regnamespace,
+  p_typname name
+)
+RETURNS boolean
+SET search_path TO 'pgtle'
+STRICT
+AS 'MODULE_PATHNAME', 'pg_tle_create_shell_type'
+LANGUAGE C;
+
+REVOKE EXECUTE ON FUNCTION pgtle.create_shell_type
+(
+  p_typnamespace regnamespace,
+  p_typname name
+) FROM PUBLIC;
+
+GRANT EXECUTE ON FUNCTION pgtle.create_shell_type
+(
+  p_typnamespace regnamespace,
+  p_typname name
+) TO pgtle_admin;
--- a/src/datatype.c
+++ b/src/datatype.c
@@ -17,60 +17,71 @@
 
 #include "access/genam.h"
 #include "access/heapam.h"
+#include "access/htup_details.h"
+#include "access/xact.h"
 #include "catalog/pg_authid.h"
 #include "catalog/pg_type.h"
+#include "fmgr.h"
 #include "miscadmin.h"
 #include "utils/acl.h"
 #include "utils/lsyscache.h"
 #include "utils/syscache.h"
 
 #include "compatibility.h"
+#include "tleextension.h"
+
+
+/* Local functions */
+static void check_is_pgtle_admin(void);
+static bool create_shell_type(Oid typeNamespace, const char *typeName, bool if_not_exists);
 
 
 static void
 check_is_pgtle_admin(void)
 {
 	HeapTuple	tuple;
-	Form_pg_authid roleform;
 	Oid			tleadminoid;
 
-	tuple = SearchSysCache1(AUTHNAME, CStringGetDatum("pgtle_admin"));
+	tuple = SearchSysCache1(AUTHNAME, CStringGetDatum(PG_TLE_ADMIN));
 	if (!HeapTupleIsValid(tuple))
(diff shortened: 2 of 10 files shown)


src/datatype.c
```diff
@@ -0,0 +1,106 @@
...
+		ereport(ERROR,
+				(errcode(ERRCODE_UNDEFINED_OBJECT),
+				 errmsg("role \"%s\" does not exist", "pgtle_admin")));
+	}
+
+	roleform = (Form_pg_authid) GETSTRUCT(tuple);
+	tleadminoid = roleform->oid;
+	ReleaseSysCache(tuple);
+
+    CHECK_CAN_SET_ROLE(GetUserId(), tleadminoid);
```
formalLogicGirl, line 49: nit : can we make the whitespace formatting consistent

test/expected/pg_tle_datatype.out
```diff
@@ -0,0 +1,64 @@
...
+(1 row)
+
+SELECT pgtle.create_shell_type('public', 'test_citext');
+ERROR:  permission denied for function create_shell_type
+-- superuser can call pgtle.create_shell_type
+RESET SESSION AUTHORIZATION;
+SELECT CURRENT_USER;
+ current_user 
+--------------
+ lyup
```
formalLogicGirl, line 31: this will not match for other people running the test and fail the test, right?

src/datatype.c
jkatz: We may need a `pgformatter` run -- it's available in the PostgreSQL repo.

src/datatype.c
```diff
@@ -0,0 +1,106 @@
...
+
+
+static void
+check_is_pgtle_admin(void)
+{
+	HeapTuple	tuple;
+	Form_pg_authid roleform;
+	Oid			tleadminoid;
+
+	tuple = SearchSysCache1(AUTHNAME, CStringGetDatum("pgtle_admin"));
```
jkatz, line 37: `pgtle_admin` should be defined in a constant. Unsure if it should be in `constants.h` or `tleextension.h` -- leaning towards `tleextension.h` as it contains names specific to the TLE extension itself, though maybe those need to be moved to `constants.h`

src/datatype.c
```diff
@@ -0,0 +1,106 @@
...
+
+	/*
+	 * Make effects of commands visible
+	 */
+	CommandCounterIncrement();
+
+	if (OidIsValid(address.objectId))
+		PG_RETURN_BOOL(true);
+
+	PG_RETURN_BOOL(false);
```
jkatz, line 105: I would opt to return an error if we're unable to create the shell type. That way, if this is in a transaction, it will fail.

src/datatype.c
```diff
@@ -0,0 +1,106 @@
...
+	typoid = GetSysCacheOid2(TYPENAMENSP, Anum_pg_type_oid,
+							 CStringGetDatum(typeName),
+							 ObjectIdGetDatum(typeNamespace));
+
+	if (OidIsValid(typoid))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_DUPLICATE_OBJECT),
+				 errmsg("type \"%s\" already exists", typeName)));
+	}
```
jkatz, line 93: See comment below -- we could have a similar pattern to other functions, e.g. `create_shell_type` which will error if the type already exists, and `create_shell_type_if_not_exists` which will return `false` if it exists. Example:

https://github.com/aws/pg_tle/blob/main/docs/03_managing_extensions.md#pgtleregister_feature_if_not_existsproc-regproc-feature-pg_tle_features

jkatz (approved):

jkatz (changes requested):
Good start! Please see previous review for the requested changes / comments.

src/datatype.c
formalLogicGirl: Since we are mirroring the Postgres CREATE TYPE wouldn't it be good to keep the behavior similar? In Postgres, if the type already exists then an error is generated. I was thinking we want to do the same here.

pg_tle--1.0.1--1.0.4.sql
```diff
@@ -107,3 +107,26 @@ GRANT EXECUTE ON FUNCTION pgtle.install_extension_version_sql
   version text,
   ext text
 ) TO pgtle_admin;
+
```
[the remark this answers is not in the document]
lyupan: Done. Added a new version 1.0.5

src/datatype.c
lyupan: Done, run `pg_indent`.

test/expected/pg_tle_datatype.out
lyupan: Ah, this was a mistake. Fixed.

src/datatype.c
lyupan: Defined a new constant `PG_TLE_ADMIN`.

src/datatype.c
lyupan: Yeah, that makes sense. Updated the code.

src/datatype.c
lyupan: Add a new create_shell_type_if_not_exists function:
`create_shell_type` now fails (report error) if the type already exists;
`create_shell_type_if_not_exists` returns false if it exists.

lyupan:
Thanks for the review! Made some updates based on the comments, please take another look.

src/datatype.c
JohnHVancouver: I think you can simplify this by just calling `get_role_oid`  and don't have to re-implement

src/datatype.c
JohnHVancouver: OOC when would a user call one over the other?

formalLogicGirl: Thats a great question. Its not clear to me that there is a very compelling use case for `create_shell_type_if_not_exists`. In Postgres, there isn't a similar function. In pg_tle, by design, we want to prevent leaving behind a shell type when a custom base type created in pg_tle gets successfully dropped. So there is not supposed to be a scenario where a shell type is left behind, and if that were the case, it seems like an error message informing the user would be helpful. Shell types also can't be shared between different custom base types, so no expectation of the shell type having been already created for some other reason. So then it comes down to whether the user wants the create function to not error under any circumstance, in which case they should use `create_shell_type_if_not_exists`. Is that a significant enough requirement that we need this function for?

formalLogicGirl: Not failing the transaction may be the significant use case that justifies the if_not_exists form of this function. Okay.

formalLogicGirl (approved):
Looks mostly good to me. I think its good to merge and work on the next functions. We can revisit if there are any updates we still need. I think you've already addressed all the comments raised so far.

PR #206, opened 2023-06-16
Implement create_base_type API.

lyupan:
Description of changes:

pgtle_admin can run create_base_type API to create a new base type (providing I/O funcs and other required arguments) in pg_tle.

User-defined input function must accept TEXT and return BYTEA, we make an C-version input function that accepts CSTRING and returns the new TYPE, the C-version input function will be used internally to create the type; inside the C-version input function, we find the user-defined input function and execute it. A dependency between C-version function and user-defined function is recorded in pg_depend, so that `DROP FUNC ... CASCADE` can work as expected.

Output function works in a similar way as the input function.

`create_base_type` implementation is similar to `DefineType` https://github.com/postgres/postgres/blob/master/src/backend/commands/typecmds.c#L148, except pg_tle specific part.

Future improvements:
1. The newly-created type can set to NULL value, and it reads/outputs NULL without invoking the user-defined I/O functions. In Postgres `CREATE TYPE` experience, input function is still called when reading a NULL input value, but it must still return NULL; output function are not invoked for NULL values. We may want to align the behavior for NULL values, but it seems to be a minor detail for now.
2. Other base type features (e.g. send/receive functions, storage, collation) are not supported in this version.


src/datatype.c
```diff
@@ -140,3 +159,809 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+	array_oid = AssignTypeArrayOid();
+
+	/*
+	 * now have TypeCreate do all the real work.
+	 *
+	 * Note: the pg_type.oid is stored in user tables as array elements (base
+	 * types) in ArrayType and in composite types in DatumTupleFields.  This
+	 * oid must be preserved by binary upgrades.
+	 */
+#if PG_VERSION_NUM >= 140000
```
jkatz, line 370: Similarly -- is there a way to put this into `compatibility.h` or the like, so there is only one call needed here?

src/datatype.c
```diff
@@ -56,7 +75,7 @@ static bool
 create_shell_type(Oid typeNamespace, const char *typeName, bool if_not_exists)
 {
 	AclResult	aclresult;
-	Oid			typoid;
+	Oid			typeoid;
```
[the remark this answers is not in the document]
lyupan: It seems both `typeOid` and `typoid` are used https://github.com/postgres/postgres/blob/0d369ac650041862ed5006885160f36d24b224a4/src/backend/commands/typecmds.c#LL1340C2-L1340C15
```
Oid			typeOid;
```

```
Oid			typoid;
```

s/typeoid/typeOid because `typeOid` seems more common

src/datatype.c
```diff
@@ -140,3 +159,809 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+PG_FUNCTION_INFO_V1(pg_tle_create_base_type);
+Datum
+pg_tle_create_base_type(PG_FUNCTION_ARGS)
+{
+	Oid			typeNamespace = PG_GETARG_OID(0);
+	char	   *typeName = NameStr(*PG_GETARG_NAME(1));
+	Oid			inputFuncId = PG_GETARG_OID(2);
+	Oid			outputFuncId = PG_GETARG_OID(3);
+	int16		internalLength = PG_GETARG_INT16(4);
+	char	   *probin = get_probin(fcinfo->flinfo->fn_oid);
```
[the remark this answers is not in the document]
lyupan: This code alignment is actually preferred by `pgindent`. (I tried to manually adjust it, but pgindent will change it back to this format)

src/datatype.c
```diff
@@ -140,3 +159,809 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+	char		alignment = TYPALIGN_INT;	/* default alignment */
+	char		storage_plain = TYPSTORAGE_PLAIN;	/* default TOAST storage
+													 * method */
+	char		storage_extended = TYPSTORAGE_EXTENDED; /* default ARRAY TOAST
+														 * storage method */
+#else
+	char		alignment = 'i';	/* default alignment */
+	char		storage_plain = 'p';	/* default TOAST storage method */
+	char		storage_extended = 'x'; /* default ARRAY TOAST storage method */
+#endif
```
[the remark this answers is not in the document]
lyupan: Good idea! Fixed.

src/datatype.c
lyupan: I don't see an elegant way to put this into `compatibility.h`. `TypeCreate(...)` is used twice, one for base type creation, one for array type creation, a few arguments are different between the two calls. Also, `TypeCreate` requires many arguments which is already error-prone ... Let me know if you have a better idea on this.

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+	 * not a way to bypass that
+	 */
+	check_is_pgtle_admin();
+
+	if (!(internalLength > 0 || internalLength == -1))
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
+				 errmsg("invalid type internal length %d", internalLength)));
+
+	if (internalLength > 0)
```
jkatz, line 248: Is this `if` necessary? Based on the last check, `internalLenth` set`

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+		internalLength += VARHDRSZ;
```
jkatz, line 249: This should provide a comment, e.g. "As a `bytea` is of variable length, we need to allow for the header"

We likely also need a check that `internalLength + VARHDRSZ` is not larger than the varlena max.

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+	 */
+#if PG_VERSION_NUM >= 120000
+	typeOid = GetSysCacheOid2(TYPENAMENSP, Anum_pg_type_oid,
+							  CStringGetDatum(typeName),
+							  ObjectIdGetDatum(typeNamespace));
+#else
+	typeOid = GetSysCacheOid2(TYPENAMENSP,
+							  CStringGetDatum(typeName),
+							  ObjectIdGetDatum(typeNamespace));
+#endif
```
jkatz, line 268: This would be easier to read as a macro in compatibility.h:

```c
#if PG_VERSION >= 120000
#define GETSYSCACHEOID2(a,b,c)  GetSysCacheOid2(a, Anum_pg_type_oid, b, c);
#else
#define GETSYSCACHEOID2(a,b,c)  GetSysCacheOid2(a, b, c);
#endif
```

Then the code can become:

```suggestion
	typeOid = GETSYSCACHEOID2(TYPENAMENSP,
							  CStringGetDatum(typeName),
							  ObjectIdGetDatum(typeNamespace));
```

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+	{
+		if (moveArrayTypeName(typeOid, typeName, typeNamespace))
+			typeOid = InvalidOid;
+		else if (if_not_exists)
+		{
+			ereport(NOTICE,
+					(errcode(ERRCODE_DUPLICATE_OBJECT),
+					 errmsg("type \"%s\" already exists, skipping", typeName)));
+			return false;
+		}
```
jkatz, line 284: If the `if_not_exists` is supposed to fail quietly, the `ereport` won't do that. It will abort the transaction. You just want to return `false`.

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+		else
+			ereport(ERROR,
+					(errcode(ERRCODE_DUPLICATE_OBJECT),
+					 errmsg("type \"%s\" already exists", typeName)));
```
jkatz, line 288: This is very if/else heavy. Here's an alternative approach:
```suggestion
		if (!moveArrayTypeName(typeOid, typeName, typeNamespace))
		{
		  if (if_not_exists)
		    return false;
		    
		  ereport(NOTICE,
					(errcode(ERRCODE_DUPLICATE_OBJECT),
					 errmsg("type \"%s\" already exists", typeName)));
		}

		typeOid = InvalidOid;
```

src/datatype.c
jkatz: I provided an example with a different function in this review. See if that helps.

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+{
+	Datum		datum = PG_GETARG_DATUM(0);
+	Oid			output_function;
+
+	output_function = find_user_output_func(get_qualified_funcname(fcinfo->flinfo->fn_oid));
+
+	/*
+	 * Call the user-defined output function.
+	 */
+	PG_RETURN_CSTRING(TextDatumGetCString(OidFunctionCall1Coll(output_function, InvalidOid, datum)));
```
jkatz, line 782: I'd break up this line, in case there are errors on the `OidFunctionCall1Coll` function.

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+							  ClanguageId,
+							  languageValidator,
+							  prosrc,	/* converted to text later */
+							  probin,	/* converted to text later */
+							  NULL, /* prosqlbody */
+							  PROKIND_FUNCTION,
+							  false,	/* security */
+							  false,	/* Leak Proof */
+							  false,	/* Strict */
+							  PROVOLATILE_IMMUTABLE,
```
jkatz, line 829: Are we validating that the user is providing an immutable function?

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+							  PROPARALLEL_SAFE,
```
jkatz, line 830: Similarly, are we validating that the user is providing a parallel safe function

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+ */
+static List *
+get_qualified_funcname(Oid funcid)
+{
+	HeapTuple	tuple;
+	Form_pg_proc proc;
+	List	   *inputNames = NIL;
+
+	tuple = SearchSysCache1(PROCOID,
+							ObjectIdGetDatum(funcid));
```
jkatz, line 947: something seems off with the indentation here.

jkatz (commented):
I think we need to ensure we have a check that the user-provided function is immutable.

src/datatype.c
lyupan: Done.

src/datatype.c
lyupan: I think so, because `internalLenth` could be -1 (variable length).

src/datatype.c
lyupan: Good catch! I don't see any size check in the postgres for `create type` command, I believe the size restriction is done by defining length as `int16`, I've added the check based on `PG_INT16_MAX`. (i.e. the length has to be le to `PG_INT16_MAX - VARHDRSZ`)

src/datatype.c
lyupan: Done. Thanks for the suggestion.

src/datatype.c
lyupan: I don't think this would abort the transaction, as the report level is `NOTICE`, similarly https://github.com/postgres/postgres/blob/master/src/backend/commands/createas.c#L421

src/datatype.c
lyupan: Done. Thanks for the suggestion.

src/datatype.c
lyupan: Done.

src/datatype.c
lyupan: Good catch. I think we actually should define this attribute based on the user input function; I've changed attributes `prosecdef,  proleakproof, proisstrict, provolatile, proparallel, procost, prorows` to match the actual user input function.

src/datatype.c
lyupan: See previous comment.

src/datatype.c
lyupan: Done.

lyupan:
> nice! Would be great to see a page for this in `pgtle/docs` with an example + comparing the `pgtle` functions with equivalent SQL commands. Its clear what's going on from the test suite, but that'll be a little harder for users to track down once merged

Hey Oliver, thanks for the suggestion! The doc is not there yet because the feature is still under development (there will be a few followup changes); once the code changes are mostly completed, we'll add a doc page with examples.

src/datatype.c
jkatz: Thanks. In some cases, we need to be more careful in the TLE interface as we're encapsulating functionality of PostgreSQL internals. It's easier for us to start out as more strict and relax later.

src/datatype.c
```diff
@@ -140,3 +153,697 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+	{
+		if (!moveArrayTypeName(typeOid, typeName, typeNamespace))
+		{
+			if (if_not_exists)
+			{
+				ereport(NOTICE,
+						(errcode(ERRCODE_DUPLICATE_OBJECT),
+						 errmsg("type \"%s\" already exists, skipping", typeName)));
+				return false;
+			}
```
jkatz, line 278: If we move the `...if_not_exists` to a PL/pgSQL function, we can just return the `ERRCODE_DUPLICATE_OBJECT` and not need this section.

src/datatype.c
```diff
@@ -140,3 +153,697 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+
+/*
+ * create_base_type
+ *
+ * Create a new base type, returns true when a new base type is successfully created.
+ *
+ * if_not_exists: if true, don't fail on duplicate name, just print a notice and return false.
+ * Otherwise, fail on duplicate name.
+ */
+static bool
```
jkatz, line 216: If we move the `if_not_exists` to a PL/pgSQL function, this can become a `static void`

src/datatype.c
```diff
@@ -140,3 +153,697 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+			ereport(ERROR,
+					(errcode(ERRCODE_DUPLICATE_OBJECT),
+					 errmsg("type \"%s\" already exists", typeName)));
+		}
+		typeOid = InvalidOid;
+	}
+
+	/*
+	 * Otherwise, we must already have a shell type, since there is no other
+	 * way that the I/O functions could have been created.
```
jkatz, line 288: Personally, when I see I/O, I think filesystem related operations, though I do understand that these are inpout/output functions.

src/datatype.c
```diff
@@ -140,3 +153,697 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+
+	/*
+	 * now have TypeCreate do all the real work.
+	 *
+	 * Note: the pg_type.oid is stored in user tables as array elements (base
+	 * types) in ArrayType and in composite types in DatumTupleFields.  This
+	 * oid must be preserved by binary upgrades.
+	 */
+	address =
+		TYPE_CREATE(false,		/* not array type */
```
jkatz, line 355: Isn't this much easier to read? 😄

src/datatype.c
```diff
@@ -140,3 +153,697 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+ * 1. must be defined in a trusted language (We check it's not in C or internal for now);
+ * 2. must accept a single argument of type text;
+ * 3. must return type bytea;
+ * 4. must be in the same namespace as the base type;
+ * 5. the to-be-created C input function must not exist yet.
+ *
+ * Raise an error if any requirement is not met.
+ */
+static void
+check_user_input_func(Oid funcid, Oid expectedNamespace)
```
jkatz, line 517: It looks like the input/output checks are almost identical, subject to a few wording changes. Can we either make the checks a common function, or refactor the code so we only have the check logic once?

For the error message wording, we can say "type function cannot ..." instead of having input/output in it. Or, have a flag that sets the appropriate language of "input" or "output"

jkatz (commented):
OK,  I think this is pretty close now. My major comment is around turning the `pgtle.create_base_type_if_not_exists` function into a PL/pgSQL function (I provided a reference to a similar one in the comments). This allows us to remove some more C code, and simplifies some of the implementation details in the other functions.

Outside of that, I think this is good to go.

src/datatype.c
```diff
@@ -140,3 +153,697 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
+				 errmsg("invalid type internal length %d", internalLength)));
+
+	if (internalLength > TLE_BASE_TYPE_SIZE_LIMIT)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
+				 errmsg("invalid type internal length %d, maximum size is %d", internalLength, TLE_BASE_TYPE_SIZE_LIMIT)));
+
+	/* As a bytea is of variable length, we need to allow for the header */
```
JohnHVancouver, line 247: Can you add comments on why/how/where you're referencing this from as well

src/datatype.c
lyupan: Yeah, they are mostly the same, updated to `check_user_defined_func` instead.

src/datatype.c
lyupan: These comments are the same as the postgres source code for create type.

pg_tle--1.0.4--1.0.5.sql
```diff
@@ -39,6 +39,34 @@ STRICT
...
+  typename name,
+  infunc regprocedure,
+  outfunc regprocedure,
+  internallength int4
+)
+RETURNS boolean
+SET search_path TO 'pgtle'
+STRICT
+AS 'MODULE_PATHNAME', 'pg_tle_create_base_type_if_not_exists'
+LANGUAGE C;
```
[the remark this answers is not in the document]
lyupan: In the latest commit, only `create_base_type_if_not_exists` is now defined in C, `create_base_type` is defined in `plpgsql` and will raise an exception if `create_base_type_if_not_exists` returns false.

The reason is that there are multiple cases where we could return `ERRCODE_DUPLICATE_OBJECT`, for example, when the shell type doesn't exist when create_baset_type is called (same code as postgres create type).

src/datatype.c
```diff
@@ -140,3 +153,697 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+	PG_RETURN_VOID();
+}
+
+/*
+ * Registers a new base type if not exists (otherwise do nothing), fails if the base type cannot be defined.
+ *
+ */
+PG_FUNCTION_INFO_V1(pg_tle_create_base_type_if_not_exists);
+Datum
+pg_tle_create_base_type_if_not_exists(PG_FUNCTION_ARGS)
```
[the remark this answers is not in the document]
lyupan: See a previous comment

src/datatype.c
lyupan: see a previous comment

src/datatype.c
lyupan: see a previous comment

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+	Oid			outputFuncId = PG_GETARG_OID(3);
+	int16		internalLength = PG_GETARG_INT16(4);
+	char	   *probin = get_probin(fcinfo->flinfo->fn_oid);
+
+	create_base_type(typeNamespace, typeName, inputFuncId, outputFuncId, internalLength, probin, false);
+
+	/*
+	 * Make effects of commands visible
+	 */
+	CommandCounterIncrement();
```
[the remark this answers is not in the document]
lyupan: On a second thought, we don't need this actually https://github.com/postgres/postgres/blob/master/src/backend/access/transam/README#L44. It is already called automatically within a transaction.

We only need it if we want to see the effect within the same command (function), for example, if we want to use one C function to do both create_shell_type and create_base_type.

src/datatype.c
```diff
@@ -140,3 +159,798 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+							  CStringGetDatum(typeName),
+							  ObjectIdGetDatum(typeNamespace));
+#else
+	typeOid = GetSysCacheOid2(TYPENAMENSP,
+							  CStringGetDatum(typeName),
+							  ObjectIdGetDatum(typeNamespace));
+#endif
+
+	/*
+	 * If it's not a shell, see if it's an autogenerated array type, and if so
```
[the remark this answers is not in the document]
lyupan: This part is the same as postgres create type implementation

https://github.com/postgres/postgres/blob/master/src/backend/commands/typecmds.c#L239
https://github.com/postgres/postgres/blob/master/src/backend/commands/typecmds.c#L606

https://www.postgresql.org/docs/current/sql-createtype.html#SQL-CREATETYPE-ARRAY

src/datatype.c
lyupan: Because we store everything as bytea which has a header, that's why we add it here.

For example, if a client wants to create a fixed length custom type int2: a 2-element vector of one byte integer value; in theory we only need two bytes (if we use `CREATE TYPE` to create it), however, because in the pg_tle implementation, it's stored using bytea, we need to account for the extra header.

pg_tle--1.0.4--1.0.5.sql
```diff
@@ -39,6 +39,42 @@ STRICT
...
+AS $_pgtleie_$
+  DECLARE
+    not_exists boolean;
+  BEGIN
+    not_exists = pgtle.create_base_type_if_not_exists(typenamespace, typename, infunc, outfunc, internallength);
+    IF NOT not_exists THEN
+      RAISE EXCEPTION 'type "%" already exists', typename;
+    END IF;
+  END;
+$_pgtleie_$
```
jkatz, line 75: I suggested the opposite -- the C function raises exceptions -- it does not return true/false. Then `pgtle.create_base_type` is a C function.

Then `pgtle.create_base_type_if_not_exists` becomes:

```sql
CREATE FUNCTION pgtle.create_base_type_if_not_exists(
  typenamespace regnamespace,
  typename name,
  infunc regprocedure,
  outfunc regprocedure,
  internallength int4
)
RETURNS bool
LANGUAGE plpgsql
AS $$
BEGIN
  PERFORM pgtle.create_base_type( typenamespace, typename, infunc, outfunc, internallength);
  RETURN TRUE;
EXCEPTION
  -- only catch the unique violation. let all other exceptions pass through.
  WHEN unique_violation THEN
    RETURN FALSE;
END;
$$;
```

jkatz (changes requested):

pg_tle--1.0.4--1.0.5.sql
lyupan: The reason I didn't choose this approach is that there are multiple cases where `create_base_type` could return `ERRCODE_DUPLICATE_OBJECT`.

For example, if `create_base_type` is called without creating shell type first, `ERRCODE_DUPLICATE_OBJECT` is raised (same error in postgres create base type without shell type); then `create_base_type_if_not_exists` would catch this error and return false (it would appear to the user as if the type already exists).

lyupan: Updated in the latest commit

jkatz (approved):

PR #209, opened 2023-06-30
Fix bug preventing TLEs auto-updating on install

### LINKED ISSUE #181: TLE installation does automatically update to the default version like Postgres extensions do
## Description

If a Postgres extension provides `foo--1.0.sql`, `foo--1.0--1.1.sql` and `foo.control` where `default_version = 1.1` then Postgres 10+ will install both scripts and follow the shortest upgrade path to update the extension to 1.1 when you run `CREATE EXTENSION foo`.

TLEs do not follow this same pattern, you must install a `foo--1.1.sql` if the default version is `1.1`.  This is the behavior of postgres prior to version 10.  For postgres 10+, the database automatically updates extensions avoiding the need for mostly duplicate files.  This prevents many existing modern extensions written in trusted languages from being usable as TLEs without adding a "dummy" duplicate version file as done here in [pgjwt](https://github.com/michelp/pgjwt/commit/b38207f3c95ac475ab89686ed9b2494428d50a4f).

## Expected outcome

As noted in [Postgres' extension packaging documentation](https://www.postgresql.org/docs/current/extend-extensions.html#id-1.8.3.20.15)

> An extension that has been around for awhile will probably exist in several versions, for which the author will need to write update scripts. For example, if you have released a foo extension in versions 1.0, 1.1, and 1.2, there should be update scripts foo--1.0--1.1.sql and foo--1.1--1.2.sql. Before PostgreSQL 10, it was necessary to also create new script files foo--1.1.sql and foo--1.2.sql that directly build the newer extension versions, or else the newer versions could not be installed directly, only by installing 1.0 and then updating. That was tedious and duplicative, but now it's unnecessary, because CREATE EXTENSION can follow update chains automatically. For example, if only the script files foo--1.0.sql, foo--1.0--1.1.sql, and foo--1.1--1.2.sql are available then a request to install version 1.2 is honored by running those three scripts in sequence. The processing is the same as if you'd first installed 1.0 and then updated to 1.2. (As with ALTER EXTENSION UPDATE, if multiple pathways are available then the shortest is preferred.) Arranging an extension's script files in this style can reduce the amount of maintenance effort needed to produce small updates.

adamguo0:
Description of changes:

An issue with how pg_tle records dependencies on the control and sql functions prevented TLEs from updating automatically via upgrade scripts to the desired version when created. This commit adds dependencies on the "base" TLE version and each necessary upgrade script.

pg_tle also did not support installing multiple versions of an extension, which has been corrected by adding "OR REPLACE" back to the control-string function (but not the sql-string function). The sql-string is executed first to detect duplicate versions being installed.

fixes #181

### DIFF
--- a/src/tleextension.c
+++ b/src/tleextension.c
@@ -2215,6 +2215,11 @@ tleCreateExtension(ParseState *pstate, CreateExtensionStmt *stmt)
 	Oid	   	   ctlfuncid = InvalidOid;
 	Oid	   	   sqlfuncid = InvalidOid;
 	ExtensionControlFile *pcontrol = NULL;
+	List	       *evi_list;
+	List           *updateVersions;
+	ExtensionVersionInfo *evi_start;
+	ExtensionVersionInfo *evi_target;
+
 
 	/* Determine if this is a pg_tle extnsion rather than a "real" extension */
 	if (strncmp(pstate->p_sourcetext, PG_TLE_MAGIC, sizeof(PG_TLE_MAGIC)) == 0)
@@ -2328,11 +2333,6 @@ tleCreateExtension(ParseState *pstate, CreateExtensionStmt *stmt)
 	 * script and all upgrade scripts.
 	 */
 
-	List	   *evi_list;
-	List       *updateVersions;
-	ExtensionVersionInfo *evi_start;
-	ExtensionVersionInfo *evi_target;
-
 	/* Extract the version update graph from the script directory */
 	evi_list = get_ext_ver_list(pcontrol);
 
@@ -2370,11 +2370,12 @@ tleCreateExtension(ParseState *pstate, CreateExtensionStmt *stmt)
 
 		foreach(lcv, updateVersions)
 		{
+			ObjectAddress upgradesqlfunc;
+
 			versionName = (char *) lfirst(lcv);
 			sqlname = psprintf("%s--%s--%s.sql", extname, oldVersionName, versionName);
 			sqlfuncid = get_tlefunc_oid_if_exists(sqlname);
 
-			ObjectAddress upgradesqlfunc;
 			upgradesqlfunc.classId = ProcedureRelationId;
 			upgradesqlfunc.objectId = sqlfuncid;
 			upgradesqlfunc.objectSubId = 0;
--- a/src/tleextension.c
+++ b/src/tleextension.c
@@ -2322,11 +2322,36 @@ tleCreateExtension(ParseState *pstate, CreateExtensionStmt *stmt)
 	if (ctlfuncid == InvalidOid)
 		elog(ERROR, "could not find control function %s for extension %s in schema %s", quote_identifier(ctlname), quote_identifier(extname), quote_identifier(PG_TLE_NSPNAME));
 
-	sqlname = psprintf("%s--%s.sql", extname, versionName);
-	sqlfuncid = get_tlefunc_oid_if_exists(sqlname);
-	if (sqlfuncid == InvalidOid)
+	/* The sql script for the installed version may not exist if the version was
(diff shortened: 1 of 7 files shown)


src/tleextension.c
```diff
@@ -2322,11 +2322,36 @@ tleCreateExtension(ParseState *pstate, CreateExtensionStmt *stmt)
...
+	/* The sql script for the installed version may not exist if the version was
+	 * installed via an upgrade script. Look for a starting version that does have
+	 * an existing install script, and then record a dependency on the install
+	 * script and all upgrade scripts.
+	 */
+
+	List	   *evi_list;
+	List       *updateVersions;
+	ExtensionVersionInfo *evi_start;
+	ExtensionVersionInfo *evi_target;
```
jkatz, line 2334: Declaring this here should trigger compiler warnings on the default PostgreSQL settings -- this should be declared at the function definition.

src/tleextension.c
```diff
@@ -2322,11 +2322,36 @@ tleCreateExtension(ParseState *pstate, CreateExtensionStmt *stmt)
...
+
+	/* Identify the target version */
+	evi_target = get_ext_ver_info(versionName, &evi_list);
+
+	/* Identify best path to reach target */
+	evi_start = find_install_path(evi_list, evi_target, &updateVersions);
+
+	/* Fail if no path ... */
+	if (evi_start == NULL)
+		elog(ERROR, "could not find sql function %s for extension %s in schema %s", quote_identifier(sqlname), quote_identifier(extname), quote_identifier(PG_TLE_NSPNAME));
```
jkatz, line 2347: Is this error only possible due to an internal state issue? If it's possible for a user to generate it, I'd suggest using `ereport` with a proper error code.

src/tleextension.c
```diff
@@ -2338,6 +2363,26 @@ tleCreateExtension(ParseState *pstate, CreateExtensionStmt *stmt)
...
+		const char *oldVersionName = versionName;
+		ListCell   *lcv;
+
+		foreach(lcv, updateVersions)
+		{
+			versionName = (char *) lfirst(lcv);
+			sqlname = psprintf("%s--%s--%s.sql", extname, oldVersionName, versionName);
+			sqlfuncid = get_tlefunc_oid_if_exists(sqlname);
+
+			ObjectAddress upgradesqlfunc;
```
jkatz, line 2377: Put this at the beginning of the `foreach` block.

src/tleextension.c
```diff
@@ -2322,11 +2322,36 @@ tleCreateExtension(ParseState *pstate, CreateExtensionStmt *stmt)
...
+	evi_start = find_install_path(evi_list, evi_target, &updateVersions);
+
+	/* Fail if no path ... */
+	if (evi_start == NULL)
+		elog(ERROR, "could not find sql function %s for extension %s in schema %s", quote_identifier(sqlname), quote_identifier(extname), quote_identifier(PG_TLE_NSPNAME));
+
+	/* Otherwise, use the best starting version */
+	versionName = evi_start->name;
 	sqlname = psprintf("%s--%s.sql", extname, versionName);
 	sqlfuncid = get_tlefunc_oid_if_exists(sqlname);
```
jkatz, line 2352: Do we still need this call here?

src/tleextension.c
adamguo0: At this point the TLE has already been created, so an upgrade path to the installed version must exist. So if an upgrade path isn't found here, it would be an internal issue, not an issue a user can generate

src/tleextension.c
adamguo0: Yes, this is to get the start version and record a dependency on its install script (before moving on to the upgrade scripts below)

jkatz (approved):

Issue #208, opened 2023-06-29
Rollback a failed transaction causes server crash


lyupan:
## Description
When `pg_tle` is installed, ROLLBACK a failed transaction causes server crash

```
BEGIN;

-- Run a query that throws ERROR
SELECT 1/0;

-- ROLLBACK would cause a server crash
ROLLBACK;

```

## Steps to reproduce
```
BEGIN;

-- Run a query that throws ERROR
SELECT 1/0;

-- ROLLBACK would cause a server crash
ROLLBACK;
```

```
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```
## Expected outcome
Without `pg_tle` installed, `ROLLBACK` would run as usual.

## Analysis
Error in the engine log

```
2023-06-29 19:09:31.558 UTC [3169] STATEMENT:  SELECT 1/0;
TRAP: FailedAssertion("IsTransactionState()", File: "relcache.c", Line: 2051, PID: 3169)
postgres: xxxx xxxx [local] ROLLBACK(ExceptionalCondition+0x90)[0x90fe00]
postgres: xxxx xxxx [local] ROLLBACK(RelationIdGetRelation+0xf7)[0x908477]
postgres: xxxx xxxx [local] ROLLBACK(relation_open+0x5b)[0x4a39bb]
postgres: xxxx xxxx [local] ROLLBACK(table_open+0x6)[0x52c856]
postgres: xxxx xxxx [local] ROLLBACK(get_extension_oid+0x23)[0x602753]
/local/home/xxxx/pg14/lib/postgresql/pg_tle.so(+0xb7f3)[0x7fc35e3a47f3]
postgres: xxxx xxxx [local] ROLLBACK[0x7f3d33]
postgres: xxxx xxxx [local] ROLLBACK[0x7f3e7a]
postgres: xxxx xxxx [local] ROLLBACK(PortalRun+0x16f)[0x7f438f]
postgres: xxxx xxxx [local] ROLLBACK[0x7efd7f]
postgres: xxxx xxxx [local] ROLLBACK(PostgresMain+0x190b)[0x7f1c3b]
postgres: xxxx xxxx [local] ROLLBACK[0x76180d]
postgres: xxxx xxxx [local] ROLLBACK(PostmasterMain+0xcbe)[0x76268e]
postgres: xxxx xxxx [local] ROLLBACK(main+0x416)[0x490776]
[...]

PR #211, opened 2023-07-03
Fix a bug that rollback a failed transaction causes server crash.

lyupan:
Fix issue #208 

The root cause of the crash is when the transaction fails, it's set to aborted state; get_extension_oid function needs a relcache lookup which must be done in a transaction state.

This change fixes this issue by skipping TransactionStmts in process utility hook.

### DIFF
--- a/src/tleextension.c
+++ b/src/tleextension.c
@@ -3977,6 +3977,27 @@ _PU_HOOK
 	Oid		tleExtensionOid;
 
 	/*
+	 * Explicitly skip TransactionStmts before calling get_extension_oid().
+	 *
+	 * In an aborted transaction, TransactionStmts (e.g. ROLLBACK) will still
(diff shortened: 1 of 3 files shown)


src/tleextension.c
```diff
@@ -3976,6 +3976,27 @@ _PU_HOOK
...
+	if (pu_parsetree && IsA(pu_parsetree, TransactionStmt))
+	{
+		if (prev_hook)
+		{
+			_prev_hook;
+		}
+		else
+		{
+			_standard_ProcessUtility;
+		}
```
jkatz, line 3996: Nonblocking comment, given it's written otherwise further down on the function, but a future refactor could simplify this to:

```suggestion
		if (prev_hook)
			_prev_hook;
		else
			_standard_ProcessUtility;
```

jkatz (approved):
Good catch and nice fix!

PR #213, opened 2023-07-11
Fix segfault when creating explicit TLE version

### LINKED ISSUE #181: TLE installation does automatically update to the default version like Postgres extensions do

adamguo0:
Description of changes:

When recording TLE dependencies for pg_dump/restore, read the extension control file whether an explicit version is specified or not.

fixes #181

### DIFF
--- a/src/tleextension.c
+++ b/src/tleextension.c
@@ -2303,10 +2303,10 @@ tleCreateExtension(ParseState *pstate, CreateExtensionStmt *stmt)
 	 * based on extension name and version
 	 */
 	extname = stmt->extname;
+	pcontrol = read_extension_control_file(extname);
 
 	if (versionName == NULL)
 	{
-		pcontrol = read_extension_control_file(extname);
 		if (pcontrol->default_version)
 			versionName = pcontrol->default_version;
 		else
--- a/test/expected/pg_tle_versions.out
+++ b/test/expected/pg_tle_versions.out
@@ -6,7 +6,9 @@
 /*
 * 1. Test that an existing version of an extension cannot be installed again
 * 2. Test that a different version of an installed extension can be installed
-* 3. Test that CREATE EXTENSION automatically updates to default version
+* 3. Test that CREATE EXTENSION with an explicit version automatically updates
+*    to that version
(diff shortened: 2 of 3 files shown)


jkatz (approved):
Good catch!

PR #217, opened 2023-07-19
Require type I/O function to be immutable and strict.

lyupan:
Although this is not strictly required in `CREATE TYPE` command, it's generally preferred. If there's a strong reason to lift the restriction in the future, we can re-consider at that time.
Also, refactored `check_user_defined_func` and `find_user_defined_func ` function a bit to improve readability.

Reference:
https://github.com/postgres/postgres/blob/master/src/backend/commands/typecmds.c#L1987

https://www.postgresql.org/docs/current/sql-createtype.html

### DIFF
--- a/src/datatype.c
+++ b/src/datatype.c
@@ -416,7 +416,7 @@ find_user_defined_func(List *procname, bool typeInput)
 		ereport(ERROR,
 				(errcode(ERRCODE_INVALID_OBJECT_DEFINITION),
 				 errmsg("type %s function %s must return type %s",
-						funcType, NameListToString(procname), format_type_be(BYTEAOID))));
+						funcType, NameListToString(procname), format_type_be(expectedRetType))));
 
 	return procOid;
 }
--- a/src/datatype.c
+++ b/src/datatype.c
@@ -388,7 +388,7 @@ find_user_defined_func(List *procname, bool typeInput)
 {
 	Oid			argList[1];
 	Oid			procOid;
-	Oid			expectedRetType;	
+	Oid			expectedRetType;
 	char	   *funcType;
 
 	/*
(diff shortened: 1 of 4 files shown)


src/datatype.c
```diff
@@ -388,13 +388,17 @@ find_user_defined_func(List *procname, bool typeInput)
...
+	char	   *funcType;
 
 	/*
 	 * User-defined input functions always take a single argument of the text
 	 * and return bytea. User-defined output functions always take a single
 	 * argument of the bytea and return text.
 	 */
 	argList[0] = typeInput ? TEXTOID : BYTEAOID;
+	expectedRetType = typeInput ? BYTEAOID : TEXTOID;
+	funcType = typeInput ? "input" : "output";
```
jkatz, line 401: It would be good to have `"input"` and `"output"` as constants somewhere.

jkatz (approved):

src/datatype.c
lyupan: Thanks for the quick review! Done.

src/datatype.c
```diff
@@ -440,52 +440,74 @@ check_user_defined_func(Oid funcid, Oid typeOid, Oid expectedNamespace, bool typ
...
+	char	   *funcType;
 
 	tuple = SearchSysCache1(PROCOID, ObjectIdGetDatum(funcid));
 	if (!HeapTupleIsValid(tuple))
 		elog(ERROR, "cache lookup failed for function %u", funcid);
 	proc = (Form_pg_proc) GETSTRUCT(tuple);
 
-	if (proc->prolang == INTERNALlanguageId || proc->prolang == ClanguageId)
+	expectedArgType = typeInput ? TEXTOID : BYTEAOID;
+	expectedRetType = typeInput ? BYTEAOID : TEXTOID;
```
adamguo0, line 459: Is it worth pulling these out into functions to avoid repeating `find_user_defined_func`?

adamguo0 (approved):

src/datatype.c
jkatz: I was thinking that -- it's probably worthwhile.

lyupan: Thanks for the suggestion! Added new functions `get_type_func_argtype` and `get_type_func_rettype`

jkatz (approved):

adamguo0 (approved):

PR #219, opened 2023-07-19
Disallow REPLACE or ALTER functions used or defined by pg_tle datatype APIs.

lyupan:
If a user-defined function is used by pgtle datatype API (i.e. `create_base_type` or `create_operator_func`), we will define a C-version function with the same name (different argument types).

Replacing or altering these functions can be error-prone. For example, if a user defined function is renamed, we won't be able to find it in the C version function because we rely on the same name as a mapping.

One note is that for operator functions, it will only be restricted if it's created or used by `create_operator_func` API.

Dropping these functions is still allowed, and will drop the dependent objects cascade.

Blocked commands:
1. `CREATE OR REPLACE FUNCTION xxx`;
2. `ALTER FUNCTION xxx`;
3. `ALTER FUNCTION xxx RENAME TO xxx`;
4. `ALTER FUNCTION xxx SET SCHEMA xxx`;
5. `ALTER FUNCTION xxx OWNER TO xxx`.

### DIFF
--- a/src/tleextension.c
+++ b/src/tleextension.c
@@ -55,6 +55,7 @@
 #endif
 #include "catalog/pg_depend.h"
 #include "catalog/pg_extension.h"
+#include "catalog/pg_language.h"
 #include "catalog/pg_namespace.h"
 #include "catalog/pg_proc.h"
 #include "catalog/pg_type.h"
@@ -71,6 +72,7 @@
 #include "nodes/pg_list.h"
 #include "nodes/plannodes.h"
 #include "parser/parse_func.h"
+#include "parser/parse_type.h"
 #include "storage/fd.h"
 #include "tcop/tcopprot.h"
 #include "tcop/utility.h"
@@ -203,6 +205,9 @@ static void pg_tle_xact_callback(XactEvent event, void *arg);
 static List *textarray_to_stringlist(ArrayType *textarray);
 static bool validate_tle_sql(char *sql);
 static void check_requires_list(List *requires);
+static bool is_pgtle_defined_c_func(Oid funcid, bool *is_operator_func);
+static bool is_pgtle_used_user_func(Oid funcid, bool *is_operator_func);
+static void check_pgtle_used_func(Oid funcid);
 
 #if PG_VERSION_NUM < 150001
 /* flag bits for InitMaterializedSRF() */
@@ -4147,6 +4152,39 @@ _PU_HOOK
 				}
 			}
 
+			/* CREATE OR REPLACE FUNCTION */
+			if (n->replace && !n->is_procedure)
+			{
+				int			nargs;
+				List	   *funcNameList;
+				Oid			funcArgList[2];
+				Oid			funcid;
+				ListCell   *x;
+				int			i = 0;
+
+				nargs = list_length(n->parameters);
+				if (nargs < 1 || nargs > 2)
+					break;
+				foreach(x, n->parameters)
+				{
+					FunctionParameter *fp = (FunctionParameter *) lfirst(x);
+					TypeName   *t = fp->argType;
+					Type		typtup = LookupTypeName(NULL, t, NULL, false);
+
+					if (!typtup)
+						ereport(ERROR,
+								(errcode(ERRCODE_UNDEFINED_OBJECT),
+									errmsg("type %s does not exist", TypeNameToString(t))));
+					funcArgList[i] = typeTypeId(typtup);
+					ReleaseSysCache(typtup);
+					++i;
+				}
+
+				funcNameList = list_make2(makeString(nspname), makeString(funcname));
+				funcid = LookupFuncName(funcNameList, nargs, funcArgList, true);
+				check_pgtle_used_func(funcid);
+			}
+
 			break;
 		}
 
@@ -4156,6 +4194,7 @@ _PU_HOOK
 			char   *funcname;
 			char   *nspname;
 			Oid		nspid;
+			Oid		funcid;			
 
 			/* Convert list of names to a name and namespace */
 			nspid = QualifiedNameGetCreationNamespace(((ObjectWithArgs *) n->func)->objname,
@@ -4195,6 +4234,9 @@ _PU_HOOK
 						errmsg("altering pg_tle functions in %s schema not allowed", PG_TLE_NSPNAME)));
 				}
 			}
+
+			funcid = LookupFuncWithArgs(n->objtype, n->func, true);
+			check_pgtle_used_func(funcid);
 
 			break;
 		}
@@ -4237,6 +4279,54 @@ _PU_HOOK
 				}
 			}
 
+			if (n->objectType == OBJECT_FUNCTION)
+			{
+				ObjectAddress address;
+				Relation	relation;
+
+				address = get_object_address(n->objectType,
+												n->object,
+												&relation,
+												AccessExclusiveLock, false);
+				check_pgtle_used_func(address.objectId);
+			}
+
+			break;
+		}
+
+		case T_RenameStmt:		/* ALTER FUNCTION xxx RENAME TO */
+		{
+			RenameStmt *stmt = (RenameStmt *) pu_parsetree;
+
+			if (stmt->renameType == OBJECT_FUNCTION)
+			{
+				ObjectAddress address;
+				Relation	relation;
+
+				address = get_object_address(stmt->renameType,
+												stmt->object,
+												&relation,
+												AccessExclusiveLock, false);
+				check_pgtle_used_func(address.objectId);
+			}
+			break;
+		}
+
+		case T_AlterOwnerStmt:	/* ALTER FUNCTION xxx OWNER TO */
+		{
+			AlterOwnerStmt *stmt = (AlterOwnerStmt *) pu_parsetree;
+
+			if (stmt->objectType == OBJECT_FUNCTION)
+			{
+				ObjectAddress address;
+				Relation	relation;
+
+				address = get_object_address(stmt->objectType,
+												stmt->object,
+												&relation,
+												AccessExclusiveLock, false);
+				check_pgtle_used_func(address.objectId);
+			}
 			break;
 		}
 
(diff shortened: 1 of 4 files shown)


src/tleextension.c
```diff
@@ -4147,6 +4152,39 @@ _PU_HOOK
...
+			{
+				int			nargs;
+				List	   *funcNameList;
+				Oid			funcArgList[2];
+				Oid			funcid;
+				ListCell   *x;
+				int			i = 0;
+
+				nargs = list_length(n->parameters);
+				if (nargs < 1 || nargs > 2)
```
formalLogicGirl, line 4166: so nargs == 1 || nargs == 2 ? Can we change this conditional to be more directly readable like that?

test/expected/pg_tle_datatype.out
```diff
@@ -323,6 +425,41 @@ SELECT pgtle.create_operator_func('public', 'test_citext', 'public.test_citext_g
...
+$$ IMMUTABLE STRICT LANGUAGE sql;
+CREATE OR REPLACE FUNCTION public.test_citext_cmp2(l bytea, r bytea)
+RETURNS int AS
+$$
+  SELECT 2;
+$$ IMMUTABLE STRICT LANGUAGE sql;
+-- ALTER type operator functions
+ALTER FUNCTION public.test_citext_cmp(l bytea, r bytea) STABLE;
+ERROR:  Alter or replace pg_tle used type I/O function test_citext_cmp is not allowed
+ALTER FUNCTION public.test_citext_cmp2(l bytea, r bytea) STABLE;
```
formalLogicGirl, line 448: shouldn't this ERROR ?

test/expected/pg_tle_datatype.out
```diff
@@ -167,6 +171,104 @@ SELECT pgtle.create_base_type('public', 'test_citext', 'test_citext_in(text)'::r
...
+-- ALTER type I/O functions
+ALTER FUNCTION public.test_citext_in(text) STABLE;
+ERROR:  Alter or replace pg_tle used type I/O function test_citext_in is not allowed
+ALTER FUNCTION public.test_citext_in(cstring) STABLE;
+ERROR:  Alter or replace pg_tle used type I/O function test_citext_in is not allowed
+ALTER FUNCTION public.test_citext_out(bytea) STABLE;
+ERROR:  Alter or replace pg_tle used type I/O function test_citext_out is not allowed
+ALTER FUNCTION public.test_citext_out(test_citext) STABLE;
+ERROR:  Alter or replace pg_tle used type I/O function test_citext_out is not allowed
+ALTER FUNCTION public.test_citext_in(int) STABLE;
```
formalLogicGirl, line 227: this shouldn't ERROR?

test/expected/pg_tle_datatype.out
```diff
@@ -167,6 +171,104 @@ SELECT pgtle.create_base_type('public', 'test_citext', 'test_citext_in(text)'::r
...
+ALTER FUNCTION public.test_citext_out(int) STABLE;
```
formalLogicGirl, line 228: or this?

test/sql/pg_tle_datatype.sql
```diff
@@ -123,6 +128,96 @@ DROP FUNCTION public.test_citext_in2;
...
+$$
+  SELECT 2;
+$$ STRICT LANGUAGE SQL;
+
+CREATE OR REPLACE FUNCTION public.another_func(input text) RETURNS bytea AS
+$$
+  SELECT pg_catalog.convert_to(input, 'UTF8');
+$$ STRICT LANGUAGE SQL;
+
+-- ALTER type I/O functions
```
formalLogicGirl, line 181: May I suggest adding some comment lines for which are positive and which are negative tests?

src/tleextension.c
jkatz: @formalLogicGirl I believe I requested it this way in a previous commit -- this was to guard against future changes to the API.

src/tleextension.c
```diff
@@ -4969,3 +5059,150 @@ static void check_requires_list(List *requires)
...
+	}
+
+	prosrcattr = SysCacheGetAttr(PROCOID, tuple,
+								 Anum_pg_proc_prosrc, &isnull);
+	Assert(!isnull);
+
+	prosrcstring = TextDatumGetCString(prosrcattr);
+	ReleaseSysCache(tuple);
+
+	return strncmp(prosrcstring, TLE_BASE_TYPE_IN, sizeof(TLE_BASE_TYPE_IN)) == 0 ||
```
rupireddAWS, line 5104: A memory leak here - TextDatumGetCString pallocs memory for prosrcstring and is_pgtle_defined_c_func() returns  without freeing it. Capture strncmp result, pfree(prosrcstring) and return the result.

src/tleextension.c
```diff
@@ -4969,3 +5059,150 @@ static void check_requires_list(List *requires)
...
+		return;
+
+	result = is_pgtle_used_user_func(funcid, &is_operator_func) ||
+			 is_pgtle_defined_c_func(funcid, &is_operator_func);
+	if (!result)
+		return;
+
+	ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					errmsg("Alter or replace pg_tle used %s function %s is not allowed",
+						   is_operator_func ? "operator" : "type I/O", get_func_name(funcid))));
```
rupireddAWS, line 5207: I don't see a test case for is_operator_func true, IOW, I couldn't find "Alter or replace pg_tle used operator function foo is not allowed" error message.

src/tleextension.c
```diff
@@ -4969,3 +5059,150 @@ static void check_requires_list(List *requires)
...
+	if (!OidIsValid(funcid))
+		return;
+
+	result = is_pgtle_used_user_func(funcid, &is_operator_func) ||
+			 is_pgtle_defined_c_func(funcid, &is_operator_func);
+	if (!result)
+		return;
+
+	ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					errmsg("Alter or replace pg_tle used %s function %s is not allowed",
```
rupireddAWS, line 5206: How about errmsg("ALTER or REPLACE of pg_tle custom datatype %s function \"%s\", is_operator_func ? "operator" : "I/O", get_func_name(funcid))));?

src/tleextension.c
```diff
@@ -4969,3 +5059,150 @@ static void check_requires_list(List *requires)
...
+
+	if (!OidIsValid(funcid))
+		return;
+
+	result = is_pgtle_used_user_func(funcid, &is_operator_func) ||
+			 is_pgtle_defined_c_func(funcid, &is_operator_func);
+	if (!result)
+		return;
+
+	ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
```
rupireddAWS, line 5205: I think it's ERRCODE_FEATURE_NOT_SUPPORTED, not ERRCODE_INVALID_PARAMETER_VALUE.

test/expected/pg_tle_datatype.out
lyupan: `ALTER FUNCTION public.test_citext_cmp2(l bytea, r bytea) STABLE;` shouldn't error because it's not an operator function used by `create_operator_func` API. I've added some comments to show which are positive and which are negative

test/expected/pg_tle_datatype.out
lyupan: This doesn't error because it's not a type input function, same function name but different argument types

test/expected/pg_tle_datatype.out
lyupan: This doesn't error because it's not a type output function, same function name but different argument types

test/sql/pg_tle_datatype.sql
lyupan: Good idea, done.

src/tleextension.c
lyupan: IIUC, palloc-ed memory will be managed by the memory context, there's no need to free them manually here

src/tleextension.c
lyupan: Good catch! There was a bug where  `is_operator_func ` is not set at all, should be fixed now

src/tleextension.c
lyupan: Updated the message to
```
	ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
					errmsg("ALTER or REPLACE of pg_tle used datatype %s function %s is not allowed",
						   is_operator_func ? "operator" : "I/O", get_func_name(funcid))));
```

src/tleextension.c
lyupan: Done

src/tleextension.c
formalLogicGirl: how does this guard against future changes to the API?

jkatz: Adding additional parameters. My opinions are loosely held on this.

lyupan: I guess `if (nargs < 1 || nargs > 2)` is a little easier to extend if we want to support for example 3+ parameters in the future. No strong opinion here either.

formalLogicGirl: ah, so extending nargs to 3, 4, etc in the future? Maybe then add a comment that nargs should be in [1,2]. But you can do that in a future PR, fine to merge this PR

formalLogicGirl (approved):

src/tleextension.c
rupireddAWS: Yes it is. But it happens only when the memory context is gone. The following is still problematic:

is_pgtle_defined_c_func is called in a loop and it leaks prosrcstring for every iteration of the loop because you're not freeing up the memory for every iteration, this loop builds up leaked memory.

```
for (; clist; clist = clist->next)
	{
		if (is_pgtle_defined_c_func(clist->oid, is_operator_func))
			return true;
	}
```

Memory contexts in postgres are poorly understood IMO, it frees up memory eventually, but when used in loops like this, the backend processes will start to consume more memory.

formalLogicGirl: if palloc/pfree work as advertised in documentation, then it would appear that developers can use palloc without explicit freeing, because the point of palloc/pfree is to reduce some of the burden of memory management. But is the concern that memory contexts themselves are not well-understood in Postgres so we can't rely on palloc/pfree as advertised?

Also, to me it seems odd that we'd have to track Postgres internal functions calls that palloc memory but that the onus of freeing memory is still on the caller?

If there is any doubt about the underlying assumptions for palloc (like memory context in postgres?) then I agree it would be safer to put in some extra explicit freeing of memory.

lyupan: The `clist` is a list of functions with a specific name `funcNameList` and two arguments, I don't think this list would go long, and it would only be used in a `ALTER FUNCTION` or `CREATE OR REPLACE FUNCTION` command. So I don't think there's a big memory pressure even if the memory is not released in every loop.

PR #221, opened 2023-07-21
Update version number 1.0.5 to 1.1.0

lyupan:
Description of changes:
Update version number 1.0.5 to 1.1.0


formalLogicGirl (approved):

rupireddAWS:
Is there any extension SQL functions or user-facing API changes that went in recently? If there are code-only changes to pg_tle, not extension SQL functions from version 1.0.5, I don't think the version needs to be updated.

It's a good practice to specify why the extension version is being updated in the commit message.

lyupan:
> Is there any extension SQL functions or user-facing API changes that went in recently? If there are code-only changes to pg_tle, not extension SQL functions from version 1.0.5, I don't think the version needs to be updated - then the customers will see new pg_tle shared library changes as part of their extension upgrades and no need to run ALTER EXTENSION ... UPDATE TO ...;
> 
Previously, we used 1.0.5 as the new version number for the data type APIs. After discussion with the team, we think it makes more sense the use 1.1.0 instead. So I created this PR to change the version number.

adamguo0:
Maybe we can include a description of the new datatype API in the commit message when the PR is merged?

adamguo0 (approved):

PR #222, opened 2023-07-24
pfree char* returned by TextDatumGetCString in is_pgtle_defined_c_func.

lyupan:
pfree char* returned by TextDatumGetCString in is_pgtle_defined_c_func.

Although this is unlikely to cause a memory leak (memory context will take care of releasing it), still be consevative and free the memory.


formalLogicGirl (approved):

lyupan:
> Was this the only place or was there a second call site?
> 
> Will merge after checks complete

I believe this is the only place we need this. There are a few TextDatumGetCString calls in datatype.c file, but they are not called in a loop, I don't think that's an issue at all.

PR #223, opened 2023-07-25
Allow superuser to replace/alter pg_tle used type functions.

lyupan:
Allow superuser to replace/alter pg_tle used type functions.
During pg_upgrade, superuser may need to replace/alter these functions.

### DIFF
--- a/src/tleextension.c
+++ b/src/tleextension.c
@@ -4153,7 +4153,7 @@ _PU_HOOK
 			}
 
 			/* CREATE OR REPLACE FUNCTION */
-			if (n->replace && !n->is_procedure)
+			if (!superuser() && n->replace && !n->is_procedure)
 			{
 				int			nargs;
 				List	   *funcNameList;
@@ -4235,8 +4235,11 @@ _PU_HOOK
 				}
 			}
 
-			funcid = LookupFuncWithArgs(n->objtype, n->func, true);
-			check_pgtle_used_func(funcid);
+			if (!superuser())
+			{
+				funcid = LookupFuncWithArgs(n->objtype, n->func, true);
+				check_pgtle_used_func(funcid);
+			}
 
 			break;
 		}
@@ -4279,7 +4282,7 @@ _PU_HOOK
 				}
 			}
 
-			if (n->objectType == OBJECT_FUNCTION)
+			if (!superuser() && n->objectType == OBJECT_FUNCTION)
 			{
 				ObjectAddress address;
 				Relation	relation;
@@ -4298,7 +4301,7 @@ _PU_HOOK
 		{
 			RenameStmt *stmt = (RenameStmt *) pu_parsetree;
 
-			if (stmt->renameType == OBJECT_FUNCTION)
+			if (!superuser() && stmt->renameType == OBJECT_FUNCTION)
 			{
 				ObjectAddress address;
 				Relation	relation;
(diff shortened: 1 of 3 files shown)


adamguo0 (approved):
With this patch, pg_upgrade succeeds on my local machine with custom types/operators created, whereas it previously failed. Thanks Lyu!

We should prioritise adding a TAP (or other) test to cover pg_upgrade.

JohnHVancouver:
Do you mind expanding in what cases pg_upgrade needs this permission?

src/tleextension.c
```diff
@@ -4153,7 +4153,7 @@ _PU_HOOK
 			}
 
 			/* CREATE OR REPLACE FUNCTION */
-			if (n->replace && !n->is_procedure)
+			if (!superuser() && n->replace && !n->is_procedure)
```
JohnHVancouver, line 4156: Would we want to use `!IsBinaryUpgrade` instead to differentiate

adamguo0:
> Do you mind expanding in what cases pg_upgrade needs this permission?

This is the specific error we are seeing in pg_upgrade with custom data types and operators:
```
 pg_restore: creating FUNCTION "public.test_citext_in("text")"
 pg_restore: creating FUNCTION "public.test_citext_in("cstring")"
 pg_restore: while PROCESSING TOC:
 pg_restore: from TOC entry 252; 1255 16435 FUNCTION test_citext_in("cstring") dbadmin
 pg_restore: error: could not execute query: ERROR:  ALTER or REPLACE of pg_tle used datatype I/O function test_citext_in is not allowed
 Command was: ALTER FUNCTION "public"."test_citext_in"("cstring") OWNER TO dbadmin;
```

src/tleextension.c
JohnHVancouver: I don't quite get the failure either. I have concerns on CREATE OR REPLACE because if you change the function definition, we think it could cause issues at the storage layer. Either corruption or reading more information. Looking at the upgrade failure that Adam provided in the github comments, it doesn't seem like we need this at all?

I'm not sure why we wouldn't want customers to redo everything from scratch and bias towards safety in the "things went wrong" case

src/tleextension.c
```diff
@@ -4298,7 +4301,7 @@ _PU_HOOK
 		{
 			RenameStmt *stmt = (RenameStmt *) pu_parsetree;
 
-			if (stmt->renameType == OBJECT_FUNCTION)
+			if (!superuser() && stmt->renameType == OBJECT_FUNCTION)
```
JohnHVancouver, line 4304: OOC what was the reason/concern behind blocking ALTER FUNCTION RENAME, ALTER FUNCTION OWNER,

lyupan: Yes

lyupan: > I don't quite get the failure either. I have concerns on CREATE OR REPLACE because if you change the function definition, we think it could cause issues at the storage layer. Either corruption or reading more information. Looking at the upgrade failure that Adam provided in the github comments, it doesn't seem like we need this at all?

> I'm not sure why we wouldn't want customers to redo everything from scratch and bias towards safety in the "things went wrong" case

We should block customers from replacing a function definition, if they did it wrong, they should redo everything.

By "things went wrong" case, I mean something went wrong from our implementation or weird cornercases. Also, IMO, allowing `superuser` to REPLACE them doesn't open any risks. For example, superuser can run `CREATE TYPE`, and they can replace the i/o functions even after the type is created.

formalLogicGirl: Are we able to guarantee (by reasoning about the code) that the customer will never have any user role that can get elevated to superuser and replace a function definition?

lyupan: AFAIK, I don't see a way for a non-superuser to get elevated to superuser

lyupan:
After discussion, https://github.com/aws/pg_tle/pull/225 was merged to fix pg_upgrade issue, this PR is not needed.

PR #227, opened 2023-08-18
Implement clientauth feature

adamguo0:
Issue #, if available:

Description of changes:

Implementation of clientauth feature, allowing users to register trusted language functions to the Postgres ClientAuthentication_hook. Uses background workers to execute database functions.


adamguo0:
I'm doing some load testing by opening a few thousand connections in parallel, and I'm encountering a race condition every few tries that results in the last one or two connections being lost.

My suspicion is that ConditionVariableSleep is not playing well with LWLocks. This is the pattern for waiting on a condition to be true:

```
while (true)
{
    LWLockAcquire(lock);
    if (condition)
    {
        LWLockRelease(lock);
        break;
    }
    // condition not met, now wait for signal
    ConditionVariableSleep(cv);
}
```

Sources that discuss condition variables (e.g. [wikipedia](https://en.wikipedia.org/wiki/Monitor_(synchronization)#Condition_variables_2)) say that LWLockRelease and ConditionVariableSleep must be atomic, or else a race condition can occur where 1. the lock is released, 2. another thread updates the state and sends the signal, and then 3. ConditionVariableSleep is called. This results in a dropped signal. I suspect this is happening because looking at the logs when the dropped connection bug happens, `clientauth_hook` goes to sleep waiting for `processed_entry` to become true, and then never wakes up.

Postgres provides a ConditionVariablePrepareToSleep method which adds a thread to the CV wait list before sleeping so I tried adding that, but the issue isn't fixed. I can't tell if ConditionVariablePrepareToSleep is supposed to handle this case, or if there's something else wrong with the logic? I would appreciate any feedback/advice, thanks!

src/clientauth.c
```diff
@@ -0,0 +1,775 @@
...
+static shmem_request_hook_type prev_shmem_request_hook = NULL;
+static void clientauth_shmem_request(void);
+#endif
+
+/* Helper functions */
+static Size clientauth_shared_memsize(void);
+static void clientauth_sighup(SIGNAL_ARGS);
+
+void clientauth_init(void);
+bool can_allow_without_executing(void);
```
JohnHVancouver, line 113: These can be static? (Other than clientauth_init)

src/clientauth.c
```diff
@@ -0,0 +1,775 @@
...
+        /* Sleep until clientauth_hook signals that a connection is ready to process */
+        while (true)
+        {
+            LWLockAcquire(clientauth_ss->lock, LW_EXCLUSIVE);
+            ConditionVariablePrepareToSleep(&clientauth_ss->bgw_process_cv);
+            idx = clientauth_ss->idx_process;
+
+            if (clientauth_ss->requests[idx].needs_processing)
+            {
+                clientauth_ss->requests[idx].needs_processing = false;
```
JohnHVancouver, line 327: I think one risk you run into with this is that if the BGW is terminated, we would have needs_processing = false even though we haven't actually processed the event. 

I imagine you're doing this to allow parallelization to happen?

src/clientauth.c
```diff
@@ -0,0 +1,775 @@
...
+    clientauth_ss->idx_insert++;
+
+    /* Signal BGW */
+    LWLockRelease(clientauth_ss->lock);
+    ConditionVariableSignal(&clientauth_ss->bgw_process_cv);
+
+    while (true)
+    {
+        LWLockAcquire(clientauth_ss->lock, LW_EXCLUSIVE);
+        ConditionVariablePrepareToSleep(&clientauth_ss->requests[idx].client_cv);
```
JohnHVancouver, line 586: OOC is there a reason we put `ConditionVariablePrepareToSleep` within the while loop instead of before?

JohnHVancouver (commented):
>  I can't tell if ConditionVariablePrepareToSleep is supposed to fix this bug or if there's something else wrong with the logic? I would appreciate any feedback/advice, thanks!

The logic seems fine. 
Can you try adding a debug line here:
https://github.com/postgres/postgres/blob/master/src/backend/storage/lmgr/condition_variable.c#L77-L79

Something like logging the MyProcPid (backend_pid) and the CV it's listening on? 

And when BGW signals add logging here:

https://github.com/postgres/postgres/blob/master/src/backend/storage/lmgr/condition_variable.c#L263-L266

To get the signal of the CV, and the PGPROC pid it signals. 

I wonder when the BGW signals the CV, if the client auth backend is actually listening on it?  
Or maybe we have multiple processes listening on the CV, and Signal only signals the oldest, while broadcast does all. 

https://github.com/postgres/postgres/blob/master/src/backend/storage/lmgr/condition_variable.c#L277-L279

Broadcast would be incorrect though cause we're only processing one entry


Also a generic comment on if the client disconnects or client auth backend terminates, do we have a stale shared memory entry perpetually?

src/clientauth.c
JohnHVancouver: Also a generic comment on if the client disconnects or client auth backend terminates, do we have a stale shared memory entry then...?

adamguo0: IIUC ConditionVariablePrepareToSleep adds the thread to CV's queue. I put it in the while loop before LWLockRelease, this way the thread is added to the wait queue before releasing the lock, preventing missed wakeups. Though maybe the thread only needs to be added once per wait loop?

src/clientauth.c
adamguo0: Sorry are you asking why they are static or saying that they should be static?

JohnHVancouver: That they should*

adamguo0:
The race condition was caused by an assumption that when `num_requests` is decremented, the next queue entry (denoted by `idx_insert`) would be available. However when a BGW is done processing, we don't know which index it was, chances are it was not the next index in queue. This resulted in entries being clobbered.

This is fixed by adding a flag to each entry denoting whether it is available. This raises two questions though:

1. A long-running query can potentially block all waiting client backends even though latter entries have become available, because client backends look sequentially through the queue
2. This parallelisation approach could cause clobbering issues elsewhere too. Per offline discussion with John we can try partitioning the queue and assigning each entry to a single BGW instead to avoid clobbering.

adamguo0:
I pushed a change that assigns each worker a partition of the queue that they are solely responsible for, instead of letting workers pick whatever entry is available for processing. This allows workers terminated mid-process to come back up and resume processing, and gives a stronger guarantee that there is no clobbering between background workers trying to process the same entry.

Rather than inserting sequentially, a client backend decides which queue entry to insert into by calculating `client_backend_pid % number_of_workers`. The PIDs should be roughly sequential which distributes the clients across the BGWs.

However in practice there is some still clustering going on. During load tests, there is a tail of clients that are waiting for the same handful of background workers to process them. We can discuss ways to avoid this or whether we should go back to the original approach.

src/clientauth.c
```diff
@@ -0,0 +1,801 @@
...
+
+            SPI_finish();
+            PopActiveSnapshot();
+            CommitTransactionCommand();
+
+            continue;
+        }
+
+        /* Get each function that is registered with clientauth */
+        proc_names = feature_proc(clientauth_feature);
```
JohnHVancouver, line 420: It's probably too much effort to bother fixing this right now. I think we could probably not have to query this table if we have a trigger on it that calls a pg_tle function to refresh feature_proc

src/clientauth.c
```diff
@@ -0,0 +1,801 @@
...
+    char    error_msg[CLIENT_AUTH_USER_ERROR_MAX_STRLEN];
+    bool    error;
+
+    if (prev_clientauth_hook)
+        prev_clientauth_hook(port, status);
+
+    /* Skip if this user is on the skip list */
+    if (check_skip_user(port->user_name))
+    {
+        elog(LOG, "%s is on pgtle.clientauth_users_to_skip, skipping", port->user_name);
```
JohnHVancouver, line 543: Not sure how I feel about LOG for these vs DEBUG1 (or logging at all) Maybe John/Jonathan has some thoughts here. It's probably useful for auditing?

src/clientauth.c
```diff
@@ -0,0 +1,801 @@
...
+        SPI_finish();
+        PopActiveSnapshot();
+        CommitTransactionCommand();
+    }
+}
+
+static void clientauth_hook(Port *port, int status)
+{
+    /* Determine the queue index that this client will insert into based on its PID */
+    int     idx = MyProc->pid % CLIENT_AUTH_MAX_PENDING_ENTRIES;
```
JohnHVancouver, line 533: Ahh you're %-ing this to get the idx to insert. I guess my earlier comments on fairness doesn't really matter too much? Well it might. 

OOC what were the trade-offs in doing this approach as opposed to something like: 

idx = MyProc % BGW and then start from some idx to try to insert into

src/clientauth.c
```diff
@@ -0,0 +1,801 @@
...
+    snprintf(clientauth_ss->requests[idx].port_info.database_name,
+             CLIENT_AUTH_PORT_SUBSET_MAX_STRLEN,
+             "%s",
+             port->database_name == NULL ? "" : port->database_name);
+    snprintf(clientauth_ss->requests[idx].port_info.user_name,
+             CLIENT_AUTH_PORT_SUBSET_MAX_STRLEN,
+             "%s",
+             port->user_name == NULL ? "" : port->user_name);
+    clientauth_ss->requests[idx].port_info.noblock = port->noblock;
+    clientauth_ss->requests[idx].port_info.remote_hostname_resolv = port->remote_hostname_resolv;
```
JohnHVancouver, line 585: I think you should use snprintf for all of these that's char[], not clear to me if you can spoof it even if there's supposed to be restrictions on length.

test/t/004_pg_tle_clientauth.pl
```diff
@@ -0,0 +1,190 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
JohnHVancouver, line 1: will review this later

JohnHVancouver (changes requested):
> We can discuss ways to avoid this or whether we should go back to the original approach.

I think using rand()  is fineif we've found that PID isn't "sequential" enough that there's still some clustering. 

Although like you mentioned I think the only time this might actually matter is when there's a connection storm such as a connection pool reconnecting and wanting to establish a bunch of new connections, or appservers re-connecting asap after a restart.  One risk we run into is if we're "unlucky" enough we lose the parallelization.

test/t/004_pg_tle_clientauth.pl
```diff
@@ -0,0 +1,190 @@
...
+$node->append_conf('postgresql.conf', qq(pgtle.enable_clientauth = 'on'));
+$node->append_conf('postgresql.conf', qq(pgtle.clientauth_num_parallel_workers = 4));
+$node->start;
+
+$node->psql('postgres', 'CREATE EXTENSION pg_tle');
+$node->psql('postgres', 'CREATE ROLE testuser LOGIN');
+
+### 1. Basic client auth function rejects invalid connections
+$node->psql('postgres', q[
+    CREATE FUNCTION reject_testuser(port pgtle.clientauth_port_subset, status integer) RETURNS text AS $$
```
JohnHVancouver, line 47: Also not sure if this should be in the actual code ie. we expect a function with these arguments, or in the docs later

src/clientauth.c
```diff
@@ -0,0 +1,801 @@
...
+        {
+            bool need_to_wake = false;
+            LWLockAcquire(clientauth_ss->lock, LW_EXCLUSIVE);
+
+            /* Check if this worker's assigned entries need processing */
+            for (int i = bgw_idx; i < CLIENT_AUTH_MAX_PENDING_ENTRIES; i += clientauth_num_parallel_workers)
+            {
+                if (clientauth_ss->requests[i].needs_processing)
+                {
+                    idx = i;
```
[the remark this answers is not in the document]
adamguo0: Yeah, one way to handle the fairness problem could be to process all of that worker's entries instead of just the first one when waking. I'm planning to try that out today

src/clientauth.c
```diff
@@ -0,0 +1,801 @@
...
+        proc_names = feature_proc(clientauth_feature);
+
+        /* Start processing the entry. First, copy the entry to local memory. */
+        LWLockAcquire(clientauth_ss->lock, LW_EXCLUSIVE);
+        memcpy(&port, &clientauth_ss->requests[idx].port_info, sizeof(port));
+        status = clientauth_ss->requests[idx].status;
+        LWLockRelease(clientauth_ss->lock);
+
+        error = false;
+        error_msg[0] = 0;
```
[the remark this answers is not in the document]
adamguo0: IIUC the integer 0 is the same as the null char \0. but \0 probably makes the intent more clear

src/clientauth.c
```diff
@@ -0,0 +1,801 @@
...
+        ConditionVariablePrepareToSleep(clientauth_ss->requests[bgw_idx].bgw_process_cv_ptr);
+        while (true)
+        {
+            bool need_to_wake = false;
+            LWLockAcquire(clientauth_ss->lock, LW_EXCLUSIVE);
+
+            /* Check if this worker's assigned entries need processing */
+            for (int i = bgw_idx; i < CLIENT_AUTH_MAX_PENDING_ENTRIES; i += clientauth_num_parallel_workers)
+            {
+                if (clientauth_ss->requests[i].needs_processing)
```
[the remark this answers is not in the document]
adamguo0: Ah yeah this is old. I meant to get rid of needs_processing after moving to the partitioned scheme. done_processing is sufficient here

Makefile
```diff
@@ -1,13 +1,13 @@
...
 SCHEMA = pgtle
 MODULE_big = $(EXTENSION)
 
-OBJS = src/tleextension.o src/guc-file.o src/feature.o src/passcheck.o src/uni_api.o src/datatype.o
+OBJS = src/tleextension.o src/guc-file.o src/feature.o src/passcheck.o src/uni_api.o src/datatype.o src/clientauth.o
 
-EXTRA_CLEAN	= src/guc-file.c pg_tle.control pg_tle--$(EXTVERSION).sql
-DATA = pg_tle.control pg_tle--1.0.0.sql pg_tle--1.0.0--1.0.1.sql pg_tle--1.0.1--1.0.4.sql pg_tle--1.0.4--1.1.1.sql pg_tle--1.1.0--1.1.1.sql
+EXTRA_CLEAN	= src/guc-file.c pg_tle.control
+DATA = pg_tle.control pg_tle--1.0.0.sql pg_tle--1.0.0--1.0.1.sql pg_tle--1.0.1--1.0.4.sql pg_tle--1.0.4.sql pg_tle--1.0.4--1.1.1.sql pg_tle--1.1.0--1.1.1.sql pg_tle--1.1.1.sql pg_tle--1.1.1--1.1.2.sql
```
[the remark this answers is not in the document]
adamguo0: When we were working on 1.0.4 and 1.1.1, the `pg_tle--$(EXTVERSION).sql` entry meant that `pg_tle--1.0.4.sql` and `pg_tle--1.1.1.sql` were getting deleted whenever I ran make clean which was inconvenient. If that's the goal though (i.e. we don't want standalone sql files for specific versions) then that makes sense

src/clientauth.c
adamguo0: I think the fairness comment still applies? The worker will still work on its first entry before moving on to other entries, even if those were filled earlier. If client 1 finishes with entry 1 and signals client 5 to insert, client 5 will get worker priority over clients 2, 3, 4 that occupied the entries 2, 3, 4.

If we put aside the fairness issue (which I think we can resolve by having the worker process all entries per wake or pick entries in a staggered way), I think the two % approaches are the same and we'll observe the same clustering behaviour.

Either the clients will pick distinct entries in the whole queue and then get folded down into BGW buckets, or they'll pick BGW buckets and then spread out across entries in the queue. In the latter approach, clients could in theory find an empty entry earlier instead of having to wait for a single predetermined entry, but since that entry is handled by the same worker anyway it would end up waiting the same amount of time (or at least in aggregate, the total wait time among clients would be the same).

src/clientauth.c
adamguo0: I think I have snprintf for all the char[] fields -- `noblock` `remote_hostname_resolv` `remote_hostname_errcode` and `status` are ints

src/clientauth.c
```diff
@@ -0,0 +1,801 @@
...
+        "",
+        PGC_SIGHUP,
+        GUC_LIST_INPUT,
+        NULL, NULL, NULL);
+
+    /* Create background workers */
+    worker.bgw_flags = BGWORKER_SHMEM_ACCESS | BGWORKER_BACKEND_DATABASE_CONNECTION;
+    worker.bgw_start_time = BgWorkerStart_RecoveryFinished;
+    worker.bgw_restart_time = 1;
+    worker.bgw_notify_pid = 0;
```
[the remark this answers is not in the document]
adamguo0: If the bgw is executing a long query, pg_ctl stop goes through right away and the client connection is closed (and the workers terminate)

src/clientauth.c
```diff
@@ -0,0 +1,801 @@
...
+{
+    int bgw_idx = DatumGetInt32(arg);
+
+    /* Establish signal handlers before unblocking signals */
+    pqsignal(SIGHUP, clientauth_sighup);
+    pqsignal(SIGTERM, die);
+    BackgroundWorkerUnblockSignals();
+
+    /* Initialize connection to the database */
+    BackgroundWorkerInitializeConnection(clientauth_database_name, NULL, 0);
```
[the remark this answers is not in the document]
adamguo0: Yeah that's right, I wonder if we could have the client backend could dynamically spin up a background worker instead of signalling one that's just waiting?

adamguo0:
Did some testing:

1. Client backend PIDs are actually more uniformly distributed than `rand()`, at least given how I run load tests (which is a lot of `psql -c 'select' &` lines)
2. Workers are not actually picking their first entry significantly more often than their other entries, it's pretty uniform. I added an offset mechanism anyway though in case this does become a problem

src/feature.c
```diff
@@ -126,6 +127,36 @@ feature_proc(const char *featurename)
 	return procs;
 }
 
+/* Check if a string is contained in a GUC parameter consisting of a comma-separated list of fields. */
+bool
+check_string_in_guc_list(const char *str, const char *guc_var, const char *guc_name)
+{
+	bool		skip = false;
```
JohnHVancouver, line 134: nit: spacing? Or is this how pg_indent does it which is weird ehh

s/skip/match ?

src/clientauth.c
```diff
@@ -0,0 +1,831 @@
...
+ * ClientAuthentication_hook.
+ *
+ * This feature uses background workers to execute the user's functions on
+ * connection attempts. The client communicates with the background workers
+ * via shared memory. Connections are written to a queue in shared memory,
+ * and a BGW is signalled to wake and process. The return value of the user's
+ * function(s) is written to the shared memory queue for the client to
+ * consume.
+ *
+ * A connection is successful if any of the following are true: 1. clientauth
```
JohnHVancouver, line 28: Maybe easier to read if it's:

```
are true:

1. ...
2.
3.
4. 
```

src/clientauth.c
```diff
@@ -0,0 +1,831 @@
...
+ * else connections will fail, since the BGW main function needs a code
+ * change to understand and pass the new struct/base type definition.
+ */
+typedef struct PortSubset
+{
+	bool		noblock;
+
+	char		remote_host[CLIENT_AUTH_PORT_SUBSET_MAX_STRLEN];
+	char		remote_hostname[CLIENT_AUTH_PORT_SUBSET_MAX_STRLEN];
+	int			remote_hostname_resolv;
```
JohnHVancouver, line 157: spacing seems a bit off

src/clientauth.c
```diff
@@ -0,0 +1,831 @@
...
+	char		database_name[CLIENT_AUTH_PORT_SUBSET_MAX_STRLEN];
+	char		user_name[CLIENT_AUTH_PORT_SUBSET_MAX_STRLEN];
+}			PortSubset;
+
+/* Represents a pending connection */
+typedef struct ClientAuthStatusEntry
+{
+	/* Data forwarded from ClientAuthentication_hook */
+	PortSubset	port_info;
+	int			status;
```
JohnHVancouver, line 169: similar comment. Maybe pgindent is worst

src/clientauth.c
```diff
@@ -0,0 +1,831 @@
...
+
+	/* Define our GUC parameters */
+	DefineCustomEnumVariable(
+							 "pgtle.enable_clientauth",
+							 gettext_noop("Sets the behavior for interacting with the pg_tle clientauth feature."),
+							 NULL,
+							 &enable_clientauth_feature,
+							 FEATURE_OFF,
+							 feature_mode_options,
+							 PGC_SIGHUP,
```
JohnHVancouver, line 266: I would probably make this PGC_POSTMASTER so you can just check below (RegisterBackgroundWorker) if you need to launch the BGWs or not and use restart to gate.  

Kinda adds friction to have to restart but I don't think we should assume people will want this feature "laying" around even if it's doing "nothing"

src/clientauth.c
```diff
@@ -0,0 +1,831 @@
...
+		LWLockAcquire(clientauth_ss->lock, LW_EXCLUSIVE);
+		clientauth_ss->requests[idx].done_processing = true;
+		clientauth_ss->requests[idx].error = error;
+		snprintf(clientauth_ss->requests[idx].error_msg, CLIENT_AUTH_USER_ERROR_MAX_STRLEN, "%s", error_msg);
+		LWLockRelease(clientauth_ss->lock);
+		ConditionVariableSignal(&clientauth_ss->requests[idx].client_cv);
+	}
+}
+
+/* Run the user's functions. This procedure should not do any transaction
```
JohnHVancouver, line 485: nit:  add line break?

src/clientauth.c
```diff
@@ -0,0 +1,831 @@
...
+		 * other workers/clients.
+		 */
+		LWLockAcquire(clientauth_ss->lock, LW_EXCLUSIVE);
+		memcpy(&port, &clientauth_ss->requests[idx].port_info, sizeof(port));
+		status = clientauth_ss->requests[idx].status;
+		LWLockRelease(clientauth_ss->lock);
+
+		/* Start a transaction in which we can run queries */
+		SetCurrentStatementStartTimestamp();
+		StartTransactionCommand();
```
JohnHVancouver, line 424: OOC do we need to do this or does SPI_exec handle that? I'm not as familiar with the StartTransaction // why we need to explicitly Begin a SubTransaction below

src/clientauth.c
```diff
@@ -0,0 +1,831 @@
...
+		/* Finish our transaction */
+		PopActiveSnapshot();
+		CommitTransactionCommand();
+
+		/*
+		 * Copy execution results back to shared memory and signal
+		 * clientauth_hook
+		 */
+		LWLockAcquire(clientauth_ss->lock, LW_EXCLUSIVE);
+		clientauth_ss->requests[idx].done_processing = true;
```
JohnHVancouver, line 477: I think you should move `done_processing` to the very end in case BGW gets terminated in between?

src/clientauth.c
```diff
@@ -0,0 +1,831 @@
...
+
+		LWLockRelease(clientauth_ss->lock);
+		ConditionVariableSleep(&clientauth_ss->requests[idx].client_cv, WAIT_EVENT_MQ_RECEIVE);
+	}
+	ConditionVariableCancelSleep();
+
+	/* Copy results of BGW processing from shared memory */
+	snprintf(error_msg, CLIENT_AUTH_USER_ERROR_MAX_STRLEN, "%s", clientauth_ss->requests[idx].error_msg);
+	error = clientauth_ss->requests[idx].error;
+	clientauth_ss->requests[idx].available_entry = true;
```
JohnHVancouver, line 682: I think I brought this up earlier in a comment. 

> Also a generic comment on if the client disconnects or client auth backend terminates, do we have a stale shared memory entry perpetually?

So I think you're fine in the case of a BGW terminates because you'll restart and work on the entry again as long as `done_processing` is not set too early. 

One issue if the ClientBackend drops/terminates is that the entry never gets cleaned up?

src/clientauth.c
JohnHVancouver: repeats elsewhere

src/feature.c
adamguo0: Ah thanks for catching that, yeah pg_indent likes to do this apparently

src/clientauth.c
adamguo0: this is pg_indent's work again.. I have to put two line breaks or else it combines the lines

src/clientauth.c
adamguo0: The subtransaction is used so we can have logic that handles SPI errors https://postgrespro.com/list/thread-id/1881982, I saw StartTransaction used in other BGW examples that do DB accesses

src/clientauth.c
adamguo0: I think pg_indent is assuming 4 character tabs but github is rendering 8 character tabs which makes it look weird

src/clientauth.c
adamguo0: Ahh that makes a lot of sense, thanks for the suggestion

adamguo0:
Pushed a fix for the stale entry bug. Copying the commit message here:
```
If a client backend terminates unexpectedly after setting
entry_available = false, it can block every subsequent client from using
that entry.

This commit adds a PID field to the request entry struct and allows
clients to check if the process currently holding the entry still
exists. If it doesn't, then it can set entry_available to true.

Signalling order is rearranged to prevent deadlocks where a backend is
terminated after entry_available is set to false but before signalling
anyone else. In addition to signalling the current client, background
workers signal the next waiting client to check if the current client
still exists.
```
I tested this by setting CLIENT_AUTH_MAX_PENDING_ENTRIES to 1, starting a connection and SIGTERMing it, then starting another connection. Previously the second connection would be blocked forever, now it goes through as expected.

src/clientauth.c
```diff
@@ -0,0 +1,868 @@
...
+	bool		error;
+
+	if (prev_clientauth_hook)
+		prev_clientauth_hook(port, status);
+
+	/* Skip if clientauth feature is off */
+	if (enable_clientauth_feature == FEATURE_OFF)
+		return;
+	/* Skip if this is a binary upgrade */
+	if (IsBinaryUpgrade)
```
JohnHVancouver, line 637: Can remove this in favor of moving IsBinaryUpgrade up to ClintAuthInit beginning

src/clientauth.c
```diff
@@ -0,0 +1,868 @@
...
+	while (true)
+	{
+		LWLockAcquire(clientauth_ss->lock, LW_EXCLUSIVE);
+
+		/*
+		 * Check if the process that's holding this entry still exists. If it
+		 * doesn't then it must have terminated uncleanly and we can set
+		 * available_entry to true.
+		 */
+		if (!BackendPidGetProc(clientauth_ss->requests[idx].pid))
```
JohnHVancouver, line 660: The reason this works is because MyBackendPid always goes to the same requests[idx] right. 

If the BackendPid has been properly authenticated, we don't need to worry about the entry ever being stale. If it doesn't exist, then the entry is by definition stale and we reset when the Client needs to use it

src/clientauth.c
```diff
@@ -0,0 +1,868 @@
...
+			break;
+
+		LWLockRelease(clientauth_ss->lock);
+		ConditionVariableSleep(clientauth_ss->requests[idx].available_entry_cv_ptr, WAIT_EVENT_MQ_RECEIVE);
+	}
+	ConditionVariableCancelSleep();
+
+	/*
+	 * Signal BGW before doing anything. This avoids a deadlock if this client
+	 * terminates after grabbing the entry but before signalling anybody.
```
JohnHVancouver, line 678: It's not clear to me why we deadlock. It seems to imply that if the backend terminates on line 674 we're still stuck which shouldn't be true?

src/clientauth.c
```diff
@@ -0,0 +1,868 @@
...
+			 */
+			snprintf(error_msg, CLIENT_AUTH_USER_ERROR_MAX_STRLEN, "%s", edata->message);
+			error = true;
+			FreeErrorData(edata);
+		}
+		PG_END_TRY();
+
+		/* Finish our transaction */
+		PopActiveSnapshot();
+		CommitTransactionCommand();
```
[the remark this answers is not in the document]
adamguo0: Yeah I think we need to be able to commit writes here. Good point on the warning, we should make it clear that these queries just run as superuser too

src/clientauth.c
adamguo0: It's not really a deadlock actually, it's more that if the backend terminates on line 674 then the BGW and the next backend will be waiting indefinitely for someone to signal them. If one of them wakes up then we can proceed as normal.

In testing it seemed like the next backend actually gets woken up when the first backend terminates? I'm not sure how

src/clientauth.c
```diff
@@ -0,0 +1,868 @@
...
+							   GUC_LIST_INPUT,
+							   NULL, NULL, NULL);
+
+	/*
+	 * If clientauth feature is enabled at postmaster startup, then register
+	 * background workers. pgtle.enable_clientauth's context is set to
+	 * PGC_POSTMASTER so that we can register background workers on postmaster
+	 * startup only if they are needed.
+	 *
+	 * Don't start the background workers if we are in pg_upgrade either.
```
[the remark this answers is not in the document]
adamguo0: Makes sense, IIUC it should be before the hooks and after the GUC inits?

src/clientauth.c
```diff
@@ -0,0 +1,868 @@
...
+								   port->database_name,
+								   port->user_name);
+
+		hookargs[0] = CStringGetTextDatum(port_subset_str);
+		hookargs[1] = Int32GetDatum(*status);
+
+		SPI_execute_with_args(query, SPI_NARGS_2, hookargtypes, hookargs, hooknulls, true, 0);
+
+		/*
+		 * Look at the first item of the first row for the return value. If
```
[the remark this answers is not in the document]
adamguo0: Working on documentation and planning to open another PR for it

src/clientauth.c
JohnHVancouver: Is this testing the case where there's only one entry? 

IIUC:

BGW loops through everything until it finds nothing to then sleeps
A new backend connection that hashes to a different bucket would always signal?

JohnHVancouver: Yea I think we are. There's a minor chance due to lock acquisition order we get a stuck backend but there is `authentication_timeout ` for those cases.

JohnHVancouver: We should update the comment fwiw

src/clientauth.c
```diff
@@ -0,0 +1,860 @@
...
+								   port->remote_hostname,
+								   port->remote_hostname_resolv,
+								   port->remote_hostname_errcode,
+								   port->database_name,
+								   port->user_name);
+
+		hookargs[0] = CStringGetTextDatum(port_subset_str);
+		hookargs[1] = Int32GetDatum(*status);
+
+		SPI_execute_with_args(query, SPI_NARGS_2, hookargtypes, hookargs, hooknulls, true, 0);
```
JohnHVancouver, line 575: I think we should check the return value here too? Might not be necessary but what do you think

https://www.postgresql.org/docs/current/spi-spi-execute-with-args.html
https://www.postgresql.org/docs/current/spi-spi-execute.html

src/clientauth.c
```diff
@@ -0,0 +1,860 @@
...
+		/*
+		 * In case the previous client terminated uncleanly, make sure the
+		 * background worker is finished with the previous entry before we
+		 * continue.
+		 */
+		if (clientauth_ss->requests[idx].available_entry && clientauth_ss->requests[idx].done_processing)
+			break;
+
+		LWLockRelease(clientauth_ss->lock);
+		ConditionVariableSleep(clientauth_ss->requests[idx].available_entry_cv_ptr, WAIT_EVENT_MESSAGE_QUEUE_RECEIVE);
```
JohnHVancouver, line 664: I guess this can potentially be race-y? However unlikely it is considering BGW needs to write a bunch of things into mem

src/clientauth.c
```diff
@@ -0,0 +1,860 @@
...
+	if (ret != SPI_OK_CONNECT)
+		ereport(ERROR,
+				(errcode(ERRCODE_CONNECTION_EXCEPTION),
+				 errmsg("\"%s.clientauth\" feature was not able to connect to the database \"%s\"",
+						PG_TLE_NSPNAME, get_database_name(MyDatabaseId))));
+
+	/*
+	 * Check if we can allow or reject without executing the user's functions
+	 */
+	if (can_allow_without_executing())
```
JohnHVancouver, line 533: If a customer has both 'require' and database_skip/user_skip configured it'll go through right. We prioritize one over the error

adamguo0: I think it makes sense for databases and users on the skip list to be allowed through if `require` is set, it can be an escape hatch if the user has locked themselves out with `require`. imo being on the skip list should mean clientauth doesn't do anything to you

src/clientauth.c
adamguo0: Do you mean if BGW signals between lines 663 and 664? IIUC `ConditionVariablePrepareToSleep` adds you to the CV queue, so if BGW signals before line 664 then ConditionVariableSleep will "wake immediately".

https://github.com/postgres/postgres/blob/8c16ad3b43299695f203f9157a2b27c22b9ed634/src/backend/storage/lmgr/condition_variable.c#L76

src/clientauth.c
adamguo0: Yeah makes sense, adding the same handling as in passcheck

JohnHVancouver: Is it always `SPI_SELECT` even if an INSERT was done by the function? 

ie. we do something like

SELECT * FROM insert_foo();

src/clientauth.c
JohnHVancouver: Wait nevermind there's the lock still which is LW_EXCLUSIVE

src/clientauth.c
adamguo0: I think it is but not sure I can guarantee that, checking < 0 also makes sense

adamguo0: Tested with INSERT and it does return SPI_OK_SELECT -- we also use the same logic in passcheck

include/compatibility.h
```diff
@@ -557,5 +557,19 @@ CastCreate(Oid sourcetypeid, Oid targettypeid, Oid funcid, char castcontext,
...
+ * PostgreSQL version 17+
+ *
+ * Renames WAIT_EVENT_MQ_* to WAIT_EVENT_MESSAGE_QUEUE_*
+ */
+#if (PG_VERSION_NUM < 170000)
+#define WAIT_EVENT_MESSAGE_QUEUE_RECEIVE WAIT_EVENT_MQ_RECEIVE
+#define WAIT_EVENT_MESSAGE_QUEUE_PUT_MESSAGE WAIT_EVENT_MQ_PUT_MESSAGE
+#define WAIT_EVENT_WAL_SENDER_WAIT_FOR_WAL WAIT_EVENT_WAL_SENDER_WAIT_WAL
+#define WAIT_EVENT_MESSAGE_QUEUE_SEND WAIT_EVENT_MQ_SEND
+#define WAIT_EVENT_MESSAGE_QUEUE_RECEIVE WAIT_EVENT_MQ_RECEIVE
```
JohnHVancouver, line 570: This one's duplicated, and we're only using one right might not need to declare all

JohnHVancouver (approved):

Issue #230, opened 2023-08-29
passcheck feature `valid_until` parameter is passing the wrong value


adamguo0:
## Description

The passcheck feature is passing an incorrect date to user functions. It seems to be passing 2000-01-01 00:00:37.560024+00.

## Steps to reproduce

1. Create and register a passcheck function that prints the `valid_until` parameter
```
psql (17devel)
Type "help" for help.

postgres=# create extension pg_tle;
CREATE EXTENSION
postgres=# create function passcheck_test(username text, password text, password_type pgtle.password_types, valid_until timestamptz, valid_null boolean)
returns void as $$
begin
raise exception '%', valid_until;
end
$$ language plpgsql;
CREATE FUNCTION
postgres=# select pgtle.register_feature('passcheck_test', 'passcheck');
 register_feature
------------------

(1 row)
```

2. Alter a role's password
```
postgres=# alter role adamguo password 'pass' valid until '2023-12-31';
ERROR:  2000-01-01 00:00:37.560024+00
CONTEXT:  PL/pgSQL function passcheck_test(text,text,pgtle.password_types,timestamp with time zone,boolean) line 3 at RAISE
SQL statement "SELECT public.passcheck_test($1::pg_catalog.text, $2::pg_catalog.text, $3::pgtle.password_types, $4::pg_catalog.timestamptz, $5::pg_catalog.bool)"
```

## Expected outcome

The error should read `ERROR:  2023-12-31 00:00:00.000000+00`, which is the given valid until date.

## Actual outcome

`ERROR:  2000-01-01 00:00:37.560024+00`, which is the wrong date.

## Analysis

In [passcheck.c](https://github.com/aws/pg_tle/blob/main/src/passcheck.c#L160), we use *GetDatum function calls to get the other four parameters, but we use DirectionFunctionCall1 to timestamptz_out for the valid_until parameter. We can try using TimestampTzGetDatum instead.

We are also missing unit tests for valid_until and valid_null.

rogerio-satler-deel:
Hi,
When running the following: 
`create user test WITH PASSWORD 'xxx' VALID UNTIL '2025-03-02';`
The value on `valid_until` variable on the parameter is `2000-09-17`
I'm still getting this same issue using the installation as:
```
SELECT pgtle.install_extension (
  'password_check_rules',
  '1.0',
  'Verify Users passwords'..
```
  Also with:
```
  SELECT pgtle.install_extension (
  'password_check_rules',
[...]

Issue #229, opened 2023-08-25
Add cluster-wide support for password check hook


adamguo0:
## Describe the problem

Currently password check hook functions need to be created and registered in each database that a user might alter passwords on. This can be difficult for admins to manage. We should allow password check functions to be created, registered, and managed from a single database.

## Describe the proposal

password_check_hook currently uses SPI to execute functions stored in the connected database. We can spin up a dynamic background worker to execute functions stored in a different database. Data can be passed between the client process and the background worker via shared memory.

## Describe alternatives

* If this feature is not built, database admins need to write scripts that manage password check functions across databases, which is a hassle.
* Using a background worker and shared memory can be somewhat resource intensive and cumbersome. It also limits parallelization, since each simultaneous background worker needs its own slice of shared memory which must be allocated at postmaster startup.

JohnHVancouver:
I think it's fine if we use the same ConditionVariable scheme you have setup for ClientAuthHook

PR #232, opened 2023-09-08
Implement cluster-wide support for passcheck hook

adamguo0:
Issue #, if available: #229 and #230 

Description of changes:

Uses a very similar approach to clientauth hook. Passcheck hook spins up dynamic background workers to query passcheck database, communicating via shared memory. The main difference is that background workers are registered dynamically when the password_check_hook is triggered -- they are terminated once the query is done. There is a single "slot" in the shared memory struct and only one background worker can exist at once.

There is potentially some refactoring that can be done to combine code between this and the clientauth hook. For example, passcheck.c registers its own shmem_startup_hook and shmem_request_hook. We can think about having unified shmem hooks if that's better, however I'm not sure how passcheck_ss and clientauth_ss would be scoped in that case. The BGW code could potentially be refactored, but the dynamic versus static approaches are a little different and the shared memory structs are different too. In summary there could be some refactoring but I'm not sure how much / if it's worth it. Perhaps if we re-use the BGW scheme for future features we'll have a better idea of how to refactor.

Introduces a new GUC, `pgtle.passcheck_db_name`. A user's existing passcheck functions will be unaffected, but functions that are not in `pgtle.passcheck_db_name` are ignored after upgrade. These functions can be re-created in `pgtle.passcheck_db_name` and registered using `pgtle.register_feature` to re-enable them. Since the background workers are dynamic, `pgtle.passcheck_db_name` is on SIGHUP context (no need to be postmaster).


adamguo0:
Rebased against HEAD

anth0nyleung:
IIUC, this commit would force passcheck to be cluster-wide. Is there a valid use case where we still want to do this per database level?

src/passcheck.c
```diff
@@ -78,111 +176,380 @@ passcheck_check_password_hook(const char *username, const char *shadow_pass, Pas
...
+				(errcode(ERRCODE_CONNECTION_EXCEPTION),
+				 errmsg("\"%s.enable_password_check\" feature was not able to connect to the database \"%s\"",
+						PG_TLE_NSPNAME, get_database_name(MyDatabaseId))));
+
+	/* Check if pg_tle extension is installed */
+	extOid = get_extension_oid(extension_name, true);
+	if (extOid == InvalidOid)
+	{
+		SPI_finish();
+		if (enable_passcheck_feature == FEATURE_REQUIRE)
```
anth0nyleung, line 438: It looks like we dont need to account for the case when feature is set to "on" anymore?

src/passcheck.c
```diff
@@ -78,111 +176,380 @@ passcheck_check_password_hook(const char *username, const char *shadow_pass, Pas
...
+/*
+ * Run the user's functions. This procedure should not do any transaction
+ * management (other than opening an SPI connection) or shared memory accesses.
+ *
+ * error and error_msg are pointers to where this procedure will store the results of function execution.
+ *
+ * passcheck_data is a pointer to the hook data that this procedure uses to execute functions.
+ */
+void
+passcheck_worker_run_user_functions(PasswordCheckHookData * passcheck_hook_data)
```
anth0nyleung, line 419: Should this be static?

src/passcheck.c
```diff
@@ -78,111 +176,380 @@ passcheck_check_password_hook(const char *username, const char *shadow_pass, Pas
...
-						errmsg("unable to execute function \"%s\"", func_name));
-		}
-		SPI_finish();
+	/* Copy data from shared memory */
+	LWLockAcquire(passcheck_ss->lock, LW_SHARED);
+	memcpy(&passcheck_hook_data, &passcheck_ss->data, sizeof(PasswordCheckHookData));
+	LWLockRelease(passcheck_ss->lock);
+
+	/* Initialize connection to the database */
+	BackgroundWorkerInitializeConnection(passcheck_database_name, NULL, 0);
```
anth0nyleung, line 333: Is it common that we assume the database exists / is valid?

anth0nyleung:
It will be beneficial if we can refactor the logic to spin up dynamic background worker and the shared memory use into something re-useable and more modular, but they is probably out of scope for this issue, maybe it's worthwile to have a different GH issue to track that (if it doesnt already exist)

adamguo0:
> IIUC, this commit would force passcheck to be cluster-wide. Is there a valid use case where we still want to do this per database level?

That's a good question, there are possibly scenarios where the database admin may want passcheck logic to be different per database. We would need to add a database parameter to the passcheck hook function for that to be possible, which would break previously defined functions.

We could potentially support both use cases (query a central database and query the current database) with a GUC toggle. Maybe that would be helpful for migrating any existing users as well

src/passcheck.c
adamguo0: Good point we should handle this case, or else the worker crashes and the user's query appears to hang

src/passcheck.c
adamguo0: The "on" handling is unchanged, if pg_tle is not installed or there are no functions registered then passcheck returns without error. If there is a function registered in pgtle.feature_info then it will execute it

src/passcheck.c
anth0nyleung: on a separate note, looks like we missed this for clientauth as well - https://github.com/aws/pg_tle/blob/main/src/clientauth.c#L353-L354C2. We can address in a different PR.

anth0nyleung:
I'd suggest checking in with @jkatz for thoughts on the transitioning from per-database to cluster-wide - https://github.com/aws/pg_tle/pull/232#issuecomment-1749070697

adamguo0:
- Decided to leave the current GUCs as is (clientauth and passcheck separate, no new worker role GUC
- Let users enable cluster-wide passcheck by setting pgtle.passcheck_db_name. If the value is empty string (which is the default), then use the legacy same-database logic

adamguo0:
> Should we also update the documentation to mention that the new GUCs we introduced, and how passcheck would be enabled cluster-wide?

Yeah definitely, I can add that into this PR or follow up with a second one

test/expected/pg_tle_api_clusterwide.out
```diff
@@ -0,0 +1,362 @@
...
+ALTER SYSTEM SET pgtle.enable_password_check = 'off';
+SELECT pg_reload_conf();
+ pg_reload_conf 
+----------------
+ t
+(1 row)
+
+-- reconnect to ensure reload settings are propagated immediately
+\c -
+ALTER ROLE testuser with password 'pass';
```
anth0nyleung, line 37: nit: Can we comment on the expected outcome of doing this?

test/t/005_pg_tle_passcheck_clusterwide.pl
```diff
@@ -0,0 +1,99 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+### 1. Function defined and registered in passcheck_db_name takes effect in other databases.
+### 2. If passcheck database does not exist, error gracefully
+### 3. If enable_passcheck = require and no functions are registered in passcheck_db_name, password cannot be set
+### 4. If passcheck function is registered in a different database, it is not called
```
anth0nyleung, line 18: Do we need a test case where
1. a passcheck function is registered is enabled only on a database
2. The passcheck function takes effective on `ALTER ROLE ... PASSWORD ...`
3. Enabled passcheck cluster wide
4. The function registered in step1 does not take effect

test/expected/pg_tle_api_clusterwide.out
```diff
@@ -0,0 +1,362 @@
...
+INSERT INTO pgtle.feature_info VALUES ('passcheck', 'testuser_2', 'unpriv_function_passcheck', 'public.unpriv_function_passcheck(pg_catalog.text,pg_catalog.text,pgtle.password_types,timestamp with time zone,boolean)');
+ERROR:  permission denied for table feature_info
+SELECT pgtle.register_feature('testuser_2.unpriv_function_passcheck', 'passcheck');
+ERROR:  permission denied for function register_feature
+SELECT pgtle.unregister_feature('password_check_length_greater_than_8', 'passcheck');
+ERROR:  permission denied for function unregister_feature
+DELETE FROM pgtle.feature_info where feature = 'passcheck';
+ERROR:  permission denied for table feature_info
+RESET SESSION AUTHORIZATION;
+CREATE OR REPLACE FUNCTION password_check_only_nums(username text, shadow_pass text, password_types pgtle.password_types, validuntil_time TimestampTz,validuntil_null boolean) RETURNS void AS
```
anth0nyleung, line 162: nit: i think we should mention here what is the intention of creating another function. Is it for testing when there are 2 functions registered for passcheck?

test/expected/pg_tle_api_clusterwide.out
```diff
@@ -0,0 +1,362 @@
...
+------------------
+ 
+(1 row)
+
+-- Test both functions are called
+ALTER ROLE testuser with password 'passwords';
+ERROR:  Passwords can only have numbers
+ALTER ROLE testuser with password '123456789';
+INSERT INTO pgtle.feature_info VALUES ('passcheck', '', 'password_check_only_nums', '');
+-- Expect to fail cause no schema qualified function found
```
anth0nyleung, line 184: nit: s/cause/because

test/expected/pg_tle_api_clusterwide.out
```diff
@@ -0,0 +1,362 @@
...
+----------------
+ t
+(1 row)
+
+DROP FUNCTION test_validuntil;
+DROP FUNCTION password_check_length_greater_than_8;
+DROP FUNCTION password_check_only_nums;
+-- OK, one more test. we're going to put a passcheck function in its own schema
+-- and then register it. we will then drop the schema and then unregister the
+-- function
```
anth0nyleung, line 257: Not blocking this PR. But generally i think it'd be good if we can refactor these different test cases in different files if possible for readability.

src/passcheck.c
```diff
@@ -78,111 +178,439 @@ passcheck_check_password_hook(const char *username, const char *shadow_pass, Pas
...
+						PG_TLE_NSPNAME, get_database_name(MyDatabaseId))));
+
+	/* Check if pg_tle extension is installed */
+	extOid = get_extension_oid(extension_name, true);
+	if (extOid == InvalidOid)
+	{
+		if (enable_passcheck_feature == FEATURE_REQUIRE)
+			ereport(ERROR,
+					errmsg("\"%s.enable_password_check\" feature is set to require but extension \"%s\" is not installed%s",
+						   PG_TLE_NSPNAME, PG_TLE_EXTNAME, database_error_msg));
```
anth0nyleung, line 502: Do we need a test for this?

test/expected/pg_tle_api_clusterwide.out
adamguo0: I agree, these tests were written pre-1.0

test/t/005_pg_tle_passcheck_clusterwide.pl
adamguo0: Added three test cases to 005_pg_tle_passcheck_clusterwide.pl

test/expected/pg_tle_api_clusterwide.out
adamguo0: Updated

src/passcheck.c
adamguo0: Added a test for this to pg_tle_api.sql and pg_tle_api_clusterwide.sql

anth0nyleung (approved):

PR #235, opened 2023-09-26
Fix validuntil time in passcheck

adamguo0:
Issue #, if available: #230

Description of changes: Use TimestampTzGetDatum to convert the validuntil_time to a Datum with the correct value.

This fix was originally in #232 but I am splitting it into its own PR so that it doesn't get held up by the other changes in that PR.

### DIFF
--- a/src/passcheck.c
+++ b/src/passcheck.c
@@ -161,7 +161,7 @@ passcheck_check_password_hook(const char *username, const char *shadow_pass, Pas
 			}
 			else
 			{
-				hookargs[3] = TimestampTzGetDatum(validuntil_time);
+				hookargs[3] = validuntil_time;
 				hookargs[4] = BoolGetDatum(false);
 			}
 
--- a/src/passcheck.c
+++ b/src/passcheck.c
@@ -161,7 +161,7 @@ passcheck_check_password_hook(const char *username, const char *shadow_pass, Pas
 			}
 			else
 			{
-				hookargs[3] = DirectFunctionCall1(timestamptz_out, validuntil_time);
+				hookargs[3] = TimestampTzGetDatum(validuntil_time);
 				hookargs[4] = BoolGetDatum(false);
 			}
 
--- a/test/expected/pg_tle_api.out
+++ b/test/expected/pg_tle_api.out
@@ -51,16 +51,31 @@ SELECT pg_reload_conf();
 ALTER ROLE testuser with password 'pass';
 ERROR:  "pgtle.enable_password_check" feature is set to require, however no entries exist in "pgtle.feature_info" with the feature "passcheck"
 -- Test validuntil_null and validuntil_time
-CREATE OR REPLACE FUNCTION test_validuntil(username text, shadow_pass text, password_types pgtle.password_types, validuntil_time TimestampTz, validuntil_null boolean) RETURNS void AS
+ALTER SYSTEM SET pgtle.enable_password_check = 'on';
+SELECT pg_reload_conf();
+ pg_reload_conf 
+----------------
+ t
+(1 row)
+
+CREATE OR REPLACE FUNCTION test_validuntil(
+    username text,
+    shadow_pass text,
+    password_types pgtle.password_types,
+    validuntil_time TimestampTz,
+    validuntil_null boolean
+) RETURNS void AS
 $$
 BEGIN
-  if validuntil_null then
+  IF validuntil_null THEN
     RAISE EXCEPTION 'Password needs a VALID UNTIL time';
-  end if;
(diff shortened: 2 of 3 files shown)


anth0nyleung:
Code change LGTM. But did we found out why using `timestamptz_out` is returning the invalid value?

adamguo0:
IIUC timestamptz_out is expecting a Datum, whereas we have the literal timestamptz value

src/passcheck.c
```diff
@@ -161,7 +161,7 @@ passcheck_check_password_hook(const char *username, const char *shadow_pass, Pas
 			}
 			else
 			{
-				hookargs[3] = DirectFunctionCall1(timestamptz_out, validuntil_time);
+				hookargs[3] = TimestampTzGetDatum(validuntil_time);
```
BRupireddy2, line 164: TimestampTzGetDatum expects TimestampTz as input, no? But the validuntil_time is a Datum passed by Postgres.

I think the following looks correct to me, do you see any problem with it?

```
TimestampTz ts = DatumGetTimestampTz(validuntil_time);
/* do whatever is wanted with ts */

/* then, to return it as a datum either do */
args[3] = TimestampTzGetDatum(ts);

/* or just return the validuntil_time if ts is unmodified */
args[3] = validuntil_time;
```

test/expected/pg_tle_api.out
```diff
@@ -50,6 +50,33 @@ SELECT pg_reload_conf();
 -- Expect an error for require if no entries are present
 ALTER ROLE testuser with password 'pass';
 ERROR:  "pgtle.enable_password_check" feature is set to require, however no entries exist in "pgtle.feature_info" with the feature "passcheck"
+-- Test validuntil_null and validuntil_time
+CREATE OR REPLACE FUNCTION test_validuntil(username text, shadow_pass text, password_types pgtle.password_types, validuntil_time TimestampTz, validuntil_null boolean) RETURNS void AS
+$$
+BEGIN
+  if validuntil_null then
```
BRupireddy2, line 57: Upper case IF ... THEN, END IF;?

test/expected/pg_tle_api.out
```diff
@@ -50,6 +50,33 @@ SELECT pg_reload_conf();
...
+    RAISE EXCEPTION 'Password needs a VALID UNTIL time';
+  end if;
+  RAISE EXCEPTION 'VALID UNTIL time: %', to_char(validuntil_time, 'YYYY-MM-DD');
```
BRupireddy2, line 60: Why raise exception for valid until time? A non-null and valid VALID UNTIL time is a positive case as far as this function is concerned, so RAISEing a NOTICE here instead of EXCEPTION looks better to me.

test/expected/pg_tle_api.out
```diff
@@ -50,6 +50,33 @@ SELECT pg_reload_conf();
 -- Expect an error for require if no entries are present
 ALTER ROLE testuser with password 'pass';
 ERROR:  "pgtle.enable_password_check" feature is set to require, however no entries exist in "pgtle.feature_info" with the feature "passcheck"
+-- Test validuntil_null and validuntil_time
+CREATE OR REPLACE FUNCTION test_validuntil(username text, shadow_pass text, password_types pgtle.password_types, validuntil_time TimestampTz, validuntil_null boolean) RETURNS void AS
```
BRupireddy2, line 54: Indentation nit-pick - the function input args can be in new lines so as to not cross 80 char.

test/sql/pg_tle_api.sql
```diff
@@ -35,6 +35,24 @@ SELECT pg_reload_conf();
...
+  if validuntil_null then
+    RAISE EXCEPTION 'Password needs a VALID UNTIL time';
+  end if;
+  RAISE EXCEPTION 'VALID UNTIL time: %', to_char(validuntil_time, 'YYYY-MM-DD');
+END;
+$$
+LANGUAGE PLPGSQL;
+SELECT pgtle.register_feature('test_validuntil', 'passcheck');
+-- Expect failure since no VALID UNTIL time is given
+ALTER ROLE testuser with password 'pass';
```
BRupireddy2, line 52: ALTER ROLE testuser with password 'pass';

Add this before registering test_validuntil as a positive case to help understand test_validuntil is in effect after pgtle.register_feature?

test/expected/pg_tle_api.out
adamguo0: Good point, thanks!

src/passcheck.c
adamguo0: Ah yeah that makes sense, for some reason I thought that validuntil_time was a TimestampTz not a Datum. The last line you posted works since we are not doing anything with the time, thank you!

anth0nyleung (approved):

PR #253, opened 2023-12-08
Convert examples to use pgtle.mk

adamguo0:
Issue #, if available:

Description of changes: Adds a `pgtle.mk` Makefile and a `create_pgtle_scripts.sh` to automatically install regular trusted language Postgres extensions using `make install`. Reorganize the examples directory so that each example is a regular Postgres extension with sql files, control file, and Makefile that can be installed with either PGXS or pg_tle Make targets.


examples/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+* `client_lockout`: Locks out database users after 5 consecutive failed logins.
+* `enforce_password_expiration`: Enforces password expiration rules.
+* `uuid_v7`: Implements UUIDv7.
+
+### Installation
+
+`pg_tle` extensions are installed by calling the functions `pgtle.install_extension` and `pgtle.install_update_path` with the extension's contents and metadata as arguments. To simplify the setup process, you can use the `pgtle.mk` Makefile to install a trusted language Postgres extension using `pg_tle` with minimal manual intervention.
+
+To install any of the examples in this directory, edit `env.ini` and specify the host, port, user, and database that you are installing to (use `PGHOST=localhost` to install locally). Then `cd` to the desired example and run `make install`. To uninstall the example, run `make uninstall`.
```
anth0nyleung, line 13: s/`cd`/change directory

examples/README.md
```diff
@@ -0,0 +1,31 @@
+## pg_tle examples
+
+This directory contains examples of `pg_tle` extensions.
+
+* `client_lockout`: Locks out database users after 5 consecutive failed logins.
+* `enforce_password_expiration`: Enforces password expiration rules.
```
anth0nyleung, line 6: I'd suggest we can move this to the README of each individual examples.

.gitignore
```diff
@@ -15,3 +15,4 @@ src/*.gcov.out
 lcov*.info
 coverage/
 coverage-html-stamp
+pgtle-*.sql
```
anth0nyleung, line 18: what is this change for?

anth0nyleung:
Is it possible to set up CI to test for installing these example?

.gitignore
adamguo0: `pgtle-{extension_name}.sql` is the build artifact generated by pgtle.mk, which we don't need to commit (`make clean` deletes it)

examples/env.ini
```diff
@@ -0,0 +1,4 @@
+PGHOST=persistent-adamguo-pgtle-hooks.cg24rccni7rd.us-west-2-elephant.rds.amazonaws.com
```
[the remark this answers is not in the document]
adamguo0: Forgot to remove, thanks!

adamguo0:
> Is it possible to set up CI to test for installing these example?

I'm trying to setup `make installcheck` in a way that lets you seamlessly run normal regression tests in a `pg_tle` context without any changes, that would be ideal to reduce friction. `pg_regress` is somewhat limiting though so I'm not sure this is possible

adamguo0:
Integrating regression tests in a seamless way will take more effort since pg_regress doesn't provide a way to run a setup script first before the tests, so let's address that in a follow up. Rebased against main, no further changes made

.gitignore
jkatz: Is there any reason we couldn't make this hidden files (e.g. `.pgtle-*.sql`) in case a user has files of such a naming convention in their folder? I'm also OK with explicitly having build files printed out if it makes more sense from a troubleshooting standpoint.

examples/README.md
```diff
@@ -0,0 +1,27 @@
+## pg_tle examples
+
+This directory contains examples of `pg_tle` extensions.
+
+### Installation
+
+`pg_tle` extensions are installed by calling the functions `pgtle.install_extension` and `pgtle.install_update_path` with the extension's contents and metadata as arguments. To simplify the setup process, you can use the `pgtle.mk` Makefile to install a trusted language Postgres extension using `pg_tle` with minimal manual intervention.
+
+To install any of the examples in this directory, you can edit `env.ini` and specify the host, port, user, and database that you are installing to (use `PGHOST=localhost` to install locally). Then you can change directory to the desired example and run `make install`. To uninstall the example, run `make uninstall`.
```
jkatz, line 9: I don't think it's a good idea to have `env.ini` as a committable file here -- that sounds like an accident waiting to happen (i.e. a user committing their DB credentials). I'd suggest:

* Change `env.ini` to `env.ini.example`
* Add `env.ini` to `.gitignore`
* Modify the instructions to ask the user to copy `env.ini.example` to `env.ini` and to update the settings as such.

examples/client_lockout/client_lockout--1.0--1.1.sql
```diff
@@ -0,0 +1,6 @@
+CREATE FUNCTION client_lockout.get_failed_attempts(user_name text)
+RETURNS int AS $$
+    SELECT num_failed_attempts
+        FROM client_lockout.failed_attempts
+        WHERE user_name = user_name
+$$ LANGUAGE sql;
```
jkatz, line 6: We should be able to set volatility to `STABLE` here as I don't think we care if this value is updated within a single transaction.

jkatz:
Forgot to mention, I think this is a really good idea -- nice work!

examples/README.md
adamguo0: Updated `pgtle-*.sql` to `.pgtle-*.sql`, and changed `env.ini` to `env.ini.example`, thanks for the suggestions!

examples/client_lockout/client_lockout--1.0--1.1.sql
adamguo0: Added

jkatz (approved):
LGTM -- Thanks!

We may ultimately want to tweak how we set the connection, but I think this works for now (and possibly into the future)!

examples/create_pgtle_scripts.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/sh
```
all2ham, line 1: why do you have `sh` shebang here and then use `[[` as the first command in the script which is incompatible

PR #256, opened 2024-01-04
Add example of ndistinct computation

samimseih:
Description of changes: Add an example of a pl/rust function that calculates and sets the n_distinct value of a column.
The calculation is performed using the HLL [1] extension.

[1] https://github.com/citusdata/postgresql-hll

### DIFF
--- a/examples/ndistinct/README.md
+++ b/examples/ndistinct/README.md
@@ -1,0 +1,115 @@
+## Calculating N-distinct elements
+
+During ```ANALYZE```, the ```n_distinct``` value of a column is calculated as either a fixed value or a ratio of the number of distinct values over the total number of rows. The fixed value is always represented as a positive number, while the ratio will be represented as a negative number. This value is visible in the n_distinct column of the ```pg_stats``` catalog view and used by the PostgreSQL query optimizer for planning. For more, refer to the PostgreSQL [[documentation]](https://www.postgresql.org/docs/current/view-pg-stats.html]).
+
+The ```ANALYZE``` command will usually calculate an optimal ```n_distinct```. However, as the table grows very large and the data has higher variability, the less likely the ```n_distinct``` value can be calculated with an optimal value. Increasing [[statistics target]](https://www.postgresql.org/docs/current/runtime-config-query.html#GUC-DEFAULT-STATISTICS-TARGET) for an ```ANALYZE``` command allows more rows and pages to be sampled, but there is a limit to how much this value can be set to.
+
+To overcome such as scenario, PostgreSQL provides users the ability to set an ```n_distinct``` value on a column to be used by ```ANALYZE```. This can be done by a DBA by using a SQL ```DISTINCT``` to calculate the number of distinct values in a column and then setting this value using the ```ALTER TABLE ... ALTER COLUMN SET (n_distinct =)``` command.
+
+This extension simplifies this task by providing a single function that calculates the n_distinct and performs the column alteration. This function can also be scheduled as part of periodic maintenance.
+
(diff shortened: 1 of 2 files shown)


examples/ndistinct/README.md
```diff
@@ -0,0 +1,135 @@
...
+To overcome such as scenario, PostgreSQL provides users the ability to set an ```n_distinct``` value on a column to be used by ```ANALYZE```. This can be done by a DBA by using a SQL ```DISTINCT``` to calculate the number of distinct values in a column and then setting this value using the ```ALTER TABLE ... ALTER COLUMN SET (n_distinct =)``` command.
+
+This extension simplifies this task by providing a single function that calculates the n_distinct and performs the column alteration. This function can also be scheduled as part of periodic maintenance.
+
+To perform the n_distinct calculation, the [[HyperLogLog]](https://github.com/citusdata/postgresql-hll) extension is used. The advantage of using ```HLL``` is that it's much more efficient than a SQL sequential scan, albeit with a small estimation error.
+
+### Prerequisites
+---
+- Install [postgresql-hll](https://github.com/citusdata/postgresql-hll)
+
```
anth0nyleung, line 16: plrust should be a pre-req as well

examples/ndistinct/ndistinct.sql
```diff
@@ -0,0 +1,248 @@
...
+                            current_ndistinct.into_datum(),
+                        ),
+                    ]
+            ))?;
+        
+        /*
+            loop through all the batches and construct the alter table... alter column
+            commands.
+        */
+        for (_, batches_rows) in all_batches.enumerate() {
```
anth0nyleung, line 167: nit: if you're not using the index, a simple for loop should be good enough

```
for rows in all_batches {
```

examples/ndistinct/README.md
samimseih: done

examples/ndistinct/ndistinct.sql
samimseih: done

examples/ndistinct/README.md
```diff
@@ -0,0 +1,115 @@
+## Calculating N-distinct elements
+
+During ```ANALYZE```, the ```n_distinct``` value of a column is calculated as either a fixed value or a ratio of the number of distinct values over the total number of rows. The fixed value is always represented as a positive number, while the ratio will be represented as a negative number. This value is visible in the n_distinct column of the ```pg_stats``` catalog view and used by the PostgreSQL query optimizer for planning. For more, refer to the PostgreSQL [[documentation]](https://www.postgresql.org/docs/current/view-pg-stats.html]).
+
+The ```ANALYZE``` command will usually calculate an optimal ```n_distinct```. However, as the table grows very large and the data has higher variability, the less likely the ```n_distinct``` value can be calculated appropriately. Increasing [[statistics target]](https://www.postgresql.org/docs/current/runtime-config-query.html#GUC-DEFAULT-STATISTICS-TARGET) for an ```ANALYZE``` command allows more rows and pages to be sampled, but there is a limit to how much this value can be set to.
```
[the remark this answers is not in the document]
samimseih: corrected the wording by removing appropriately

examples/ndistinct/ndistinct.sql
```diff
@@ -0,0 +1,251 @@
...
+                            current_ndistinct_all[i],
+                            alter_sql);
+                }
+            }
+        }
+        Ok(None)
+    }
+    )
+$$ LANGUAGE plrust STRICT VOLATILE;
+$_pg_tle_$
```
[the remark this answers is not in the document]
samimseih: added this but it needs to an array such as 

```
$_pg_tle_$,
ARRAY['plrust', 'hll']
```

PR #257, opened 2024-01-05
Erase shared memory entry after each clientauth/passcheck request

adamguo0:
Issue #, if available:

Description of changes: After each clientauth and passcheck request is completed by the background worker, erase the shared memory entry to prevent sensitive information from persisting longer than necessary.

### DIFF
--- a/src/clientauth.c
+++ b/src/clientauth.c
@@ -737,6 +737,7 @@ clientauth_hook(Port *port, int status)
 	/* Copy results of BGW processing from shared memory */
 	snprintf(error_msg, CLIENT_AUTH_USER_ERROR_MAX_STRLEN, "%s", clientauth_ss->requests[idx].error_msg);
 	error = clientauth_ss->requests[idx].error;
+
 	/* Erase data about this request from shared memory now that we're done */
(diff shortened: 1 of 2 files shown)


src/clientauth.c
```diff
@@ -737,6 +737,12 @@ clientauth_hook(Port *port, int status)
 	/* Copy results of BGW processing from shared memory */
 	snprintf(error_msg, CLIENT_AUTH_USER_ERROR_MAX_STRLEN, "%s", clientauth_ss->requests[idx].error_msg);
 	error = clientauth_ss->requests[idx].error;
+	/* Erase data about this request from shared memory now that we're done */
```
JohnHVancouver, line 740: Maybe add a line break for readability before this. 
Is this to ideally prevent keeping things around in shared memory? I was under the impression we're roughly guaranteed to reset the shared memory entry when we process the next request?

https://github.com/aws/pg_tle/blob/main/src/clientauth.c#L510-L511

adamguo0: Yeah that's right, but until the next request comes in the data will hang around in shared memory.

JohnHVancouver (approved):

PR #259, opened 2024-01-12
Let Postgres start if clientauth hits an error

adamguo0:
Issue #, if available:

Description of changes: Currently we block Postgres from starting if clientauth cannot initialize its background workers. This causes more downtime and headache than necessary. Instead, use a flag to track whether clientauth hit an error initializing and effectively disable the feature if so, emitting warnings to the database log.

### DIFF
--- a/src/clientauth.c
+++ b/src/clientauth.c
@@ -140,6 +140,7 @@ static char *clientauth_databases_to_skip = "";
 
 /* Global flags */
 static bool clientauth_reload_config = false;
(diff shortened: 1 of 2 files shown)


JohnHVancouver:
More generally, what's the right behavior? If a user wants to start with client_auth hook and they're relying on this feature to authenticate logins, is there a reason they would want the database to start as opposed to fixing the issue?

adamguo0:
Good point, closing this PR

PR #260, opened 2024-01-12
Add author attribution to ndistinct README

samimseih:
Description of changes:
Adding missing attribution to the ndistinct README


jkatz:
Have we done this for other examples?

adamguo0:
We haven't -- it's not too relevant for our existing examples since to my knowledge they weren't proposed or discussed elsewhere first (though I could be wrong about uuid_v7).

jkatz:
If we go this route, I think for UUIDv7 we should reference the upstream PostgreSQL thread as this predates the implementation and did inspire our discussion around adding it: https://www.postgresql.org/message-id/CAAhFRxitJv%3DyoGnXUgeLB_O%2BM7J2BJAmb5jqAT9gZ3bij3uLDA%40mail.gmail.com

adamguo0:
Makes sense, I can add that to this PR

Issue #263, opened 2024-01-23
size limitation on custom type.


thiru-baffle:
## Description

we have created the custom type. And attempting to update/insert data to the table of custom column via COPY command.  
PSQLException: ERROR: row is too big: size 1088976, maximum size 8160

## Steps to reproduce

1. created the custom type.
2.  create a table using custom type as a column.
3. COPY TO export the file to the table of the custom column. 
4. The data should be of size ~1 MB.

## Expected outcome

data should be inserted properly.

## Actual outcome

PSQLException: ERROR: row is too big: size 1088976, maximum size 8160

## Analysis
same data can be converted from text -> bytea using the 'infunc' of the custom type.

If applicable, please provide logs that demonstrate the issue.
Please remove any sensitive information from the logs.

adamguo0:
Hi @thiru-baffle, can you post some sample code to reproduce the issue?

adamguo0:
I believe we would need custom datatypes to support TOAST for this to work: https://www.postgresql.org/docs/current/storage-toast.html
```
Only certain data types support TOAST — there is no need to impose the overhead on data types that cannot produce large field values. To support TOAST, a data type must have a variable-length (varlena) representation, in which, ordinarily, the first four-byte word of any stored value contains the total length of the value in bytes (including itself). TOAST does not constrain the rest of the data type's representation.
```

thiru-baffle:
create extension if not exists pgcrypto;
create extension if not exists citext;
create schema if not exists baffle_udfs;
create extension if not exists pg_tle;

drop function if exists baffle_udfs.bafflectr2_in(input text) cascade;
create or replace function baffle_udfs.bafflectr2_in(input text)
  returns bytea
  immutable strict language 'plpgsql' as
$$
  declare hex_text text;
  declare hex_bytes bytea;
begin
  if starts_with(input, '\x') then
    hex_text := substr(input, 3);
  else
    hex_text := input;
  end if;
  hex_bytes := decode(hex_text, 'hex');
  return hex_bytes;
end;
$$;

drop function if exists baffle_udfs.bafflectr2_out(input bytea) cascade;
create or replace function baffle_udfs.bafflectr2_out(input bytea)
  returns text
  immutable strict language 'plpgsql' as
$$
begin
  return '\x' || encode(input, 'hex');
end;
$$;


do
$anonfn0$
begin
  perform pgtle.create_shell_type('baffle_udfs', 'bafflectr2');
[...]

PR #266, opened 2024-02-07
Support customized alignment and storage for pg_tle create type API

lyupan:
Issue #263.

In the previous version, we used default int4 alignment and plain storage (non-TOASTable). The data will always be stored in-line and not compressed. An error will be reported if the value exceeds the limit:
`ERROR: row is too big: size xxx, maximum size xxx`

With this change, users can customize the alignment and storage strategies. Compression or out-of-line storage can be used depending on the storage strategy (https://www.postgresql.org/docs/current/storage-toast.html).

The default alignment is int4 while the default storage is plain, compatible with previous versions.

A new version 1.3.5 is created along with this change.


src/datatype.c
```diff
@@ -159,6 +168,57 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+/*
+ * pg_tle_create_base_type_7args
+ *
+ * Similar to pg_tle_create_base_type, but 7 args are expected.
+ * Two additional args are alignment and storage.
+ *
+ */
+PG_FUNCTION_INFO_V1(pg_tle_create_base_type_7args);
+Datum
+pg_tle_create_base_type_7args(PG_FUNCTION_ARGS)
```
adamguo0, line 193: nit: is adding `7args` to the name standard practice? Maybe `pg_tle_create_base_type_with_storage` would be more descriptive

Makefile
```diff
@@ -1,13 +1,13 @@
 EXTENSION = pg_tle
-EXTVERSION = 1.3.4
+EXTVERSION = 1.3.5
```
adamguo0, line 2: Let's get some opinions on whether to bump the version to 1.3.5 or 1.4.0. My initial thought was 1.3.5 since this is a small update to an existing feature, but it also introduces an API change which may warrant a minor version bump.

src/datatype.c
lyupan: I'm not sure if this is the standard practice, but I have seen this naming style in postgres source code, for example
`gin_extract_tsquery_5args(PG_FUNCTION_ARGS)` https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/tsginidx.c#L318

docs/09_datatypes.md
```diff
@@ -63,6 +63,8 @@ SELECT pgtle.create_shell_type_if_not_exists('public', 'test_citext');
...
 * `outfunc`: The name of a previously defined function to convert the type's internal binary representation (`bytea`) to the external textual representation. The function must take one argument of type `bytea` and return `text`. The function must also be declared as `IMMUTABLE` and `STRICT`. It will not be called with a NULL parameter.
 * `internallength`: Total length of the base data type as bytes. Base data types can be fixed-length, in which case internallength is a positive integer, or variable-length, in which case internallength is -1.
+* `alignment`: The optional alignment parameter specifies the storage alignment requirement of the data type. Allowed values are 'char', 'int2', 'int4', 'double'. The allowed values equate to alignment on 1, 2, 4, or 8 byte boundaries. The default is 'int4', alignment on 4 byte boundaries. Note that variable-length types must have an alignment of at least 4, since they necessarily contain an int4 as their first component.
```
anth0nyleung, line 66: Just a sanity check, is the allowed values that same for all major version we support?

anth0nyleung:
could you also test upgrading to this version works and share the results?

Makefile
jkatz: If we go strictly by semantic versioning, a new, backwards compatible change would require a minor version bump (`1.4.0`)

docs/09_datatypes.md
```diff
@@ -48,7 +48,7 @@ SELECT pgtle.create_shell_type('public', 'test_citext');
 SELECT pgtle.create_shell_type_if_not_exists('public', 'test_citext');
 ```
 
-### `pgtle.create_base_type(typenamespace regnamespace, typename name, infunc regprocedure, outfunc regprocedure, internallength int4)`
+### `pgtle.create_base_type(typenamespace regnamespace, typename name, infunc regprocedure, outfunc regprocedure, internallength int4, alignment text default 'int4', storage text default 'plain')`
```
jkatz, line 51: Why default to `plain` and not `extended`? Using a toastable storage type would eliminate "gotchas" if builders create types that exceed the TOAST threshold.

docs/09_datatypes.md
jkatz: I'd recommend giving more guidance on how this is used. Most builders are not going to be familiar with byte alignment.

> Note that variable-length types must have an alignment of at least 4, since they necessarily contain an int4 as their first component.

I'd recommend rephrasing this:

> A variable-length type (e.g., `text`) must have a byte-alignment of at least `4` due to the size of its header.

docs/09_datatypes.md
```diff
@@ -96,6 +102,8 @@ SELECT pgtle.create_base_type('public', 'test_int2', 'test_int2_in(text)'::regpr
 SELECT pgtle.create_base_type_if_not_exists('public', 'test_citext', 'test_citext_in(text)'::regprocedure, 'test_citext_out(bytea)'::regprocedure, -1);
 -- Create a fixed-length (2 bytes) data type
 SELECT pgtle.create_base_type_if_not_exists('public', 'test_int2', 'test_int2_in(text)'::regprocedure, 'test_int2_out(bytea)'::regprocedure, 2);
+-- Create a TOASTable variable-length data type 
+SELECT pgtle.create_base_type_if_not_exists('public', 'test_citext', 'test_citext_in(text)'::regprocedure, 'test_citext_out(bytea)'::regprocedure, -1, storage => 'extended');
```
jkatz, line 106: Note if we change the default to `extended` this should also be updated.

src/datatype.c
jkatz: I'm not a fan of this naming scheme. While there's precedent in upstream, we don't have to mimic it from a user-exposed interface.

+1 to @adamguo0 suggestion: `pg_tle_create_base_type_with_storage`, or maybe `pg_tle_create_base_type_with_storage_type`

jkatz: Also - to expose this, would we be able to rewrite the existing function to support 7 arguments, and have defaults fo rthe ones that are not necessarily set?

src/datatype.c
```diff
@@ -201,6 +257,32 @@ pg_tle_create_base_type(PG_FUNCTION_ARGS)
...
+	else if (pg_strcasecmp(alignmentStr, "int4") == 0)
+		alignment = TYPALIGN_INT;
+	else if (pg_strcasecmp(alignmentStr, "int2") == 0)
+		alignment = TYPALIGN_SHORT;
+	else if (pg_strcasecmp(alignmentStr, "char") == 0)
+		alignment = TYPALIGN_CHAR;
+	else
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("alignment \"%s\" not recognized", alignmentStr)));
```
jkatz, line 271: Suggestion: put this lookup into its own function.

src/datatype.c
```diff
@@ -201,6 +257,32 @@ pg_tle_create_base_type(PG_FUNCTION_ARGS)
...
+	else if (pg_strcasecmp(storageStr, "external") == 0)
+		storage = TYPSTORAGE_EXTERNAL;
+	else if (pg_strcasecmp(storageStr, "extended") == 0)
+		storage = TYPSTORAGE_EXTENDED;
+	else if (pg_strcasecmp(storageStr, "main") == 0)
+		storage = TYPSTORAGE_MAIN;
+	else
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("storage \"%s\" not recognized", storageStr)));
```
jkatz, line 284: Suggestion: put this lookup into its own function.

src/datatype.c
lyupan: Updated function name to `pg_tle_create_base_type_with_storage`. 

The C function `pg_tle_create_base_type_7args/with_storage` is not exposed to the users, only `pg_tle_create_base_type` function (used in the previous 1.3.4 version) is exposed to the user.
```
v1.3.4: create function pg_tle.create_base_type() .... AS 'pg_tle_create_base_type'
v1.4.0+: create function pg_tle.create_base_type() .... AS 'pg_tle_create_base_type_with_storage'
```
Adding an internal `pg_tle_create_base_type_with_storage` C function (without modifying the exist one) makes it cleaner and easier to maintain.

Background: in my first commit, I modified the `pg_tle_create_base_type` C function to add these 2 arguments, however, it has a hidden bug for upgrade (See commit message in https://github.com/aws/pg_tle/pull/266/commits/f56985ba9b9710f0de09105de05a80f5e3dff1dd), this is discovered by Josh Tiefenbach and I made this update based on the suggestion, we think this approach creates less confusion and maintenance issues.

Makefile
lyupan: Thanks for the suggestion. Updated the new version number to `1.4.0`

docs/09_datatypes.md
lyupan: > Just a sanity check, are the allowed values the same for all major version we support?
Yes, this didn't change at least since PG11, https://www.postgresql.org/docs/11/sql-createtype.html
```
alignment
The storage alignment requirement of the data type. If specified, it must be char, int2, int4, or double; the default is int4.

storage
The storage strategy for the data type. If specified, must be plain, external, extended, or main; the default is plain.
```

lyupan:
> could you also test upgrading to this version works and share the results?
Yes

1. Create pg_tle with old version library and version
```
postgres=# create extension pg_tle;
CREATE EXTENSION
postgres=# \dx
                        List of installed extensions
  Name   | Version |   Schema   |                Description                 
---------+---------+------------+--------------------------------------------
 pg_tle  | 1.3.4   | pgtle      | Trusted Language Extensions for PostgreSQL
 plpgsql | 1.0     | pg_catalog | PL/pgSQL procedural language
(2 rows)
```

2. Create custom types and data
```
...
-- Variable-length
postgres=# SELECT pgtle.create_base_type('public', 'test_citext', 'test_citext_in(text)'::regprocedure, 'test_citext_out(bytea)'::regprocedure, -1);
 create_base_type 
------------------
 
(1 row)

-- fixed length
postgres=# SELECT pgtle.create_base_type('public', 'test_int2', 'test_int2_in(text)'::regprocedure, 'test_int2_out(bytea)'::regprocedure, 2);
 create_base_type 
------------------
 
(1 row)
postgres=# INSERT INTO test_dt_int2 VALUES ('11,22');
INSERT 0 1
postgres=# SELECT * FROM test_dt_int2;
  c1   
-------
 11,22
(1 row)

postgres=# INSERT INTO test_dt VALUES ('SELECT'), ('INSERT'), ('UPDATE'), ('DELETE');
INSERT 0 4
postgres=# select * from test_dt;
   c1   
--------
 SELECT
 INSERT
 UPDATE
 DELETE
(4 rows)

```

3. Install the new extension library (1.4.0), restart postgres and upgrade
```
postgres=# ALTER EXTENSION pg_tle UPDATE;
ALTER EXTENSION
postgres=# \dx
                        List of installed extensions
  Name   | Version |   Schema   |                Description                 
---------+---------+------------+--------------------------------------------
 pg_tle  | 1.4.0   | pgtle      | Trusted Language Extensions for PostgreSQL
 plpgsql | 1.0     | pg_catalog | PL/pgSQL procedural language
(2 rows)
```

4. Check the old types (tables) can still read/write
```
postgres=# select * from test_dt;
   c1   
--------
 SELECT
 INSERT
 UPDATE
 DELETE
(4 rows)

postgres=# select * from test_dt_int2;
  c1   
-------
 11,22
(1 row)

postgres=# delete from test_dt;
DELETE 4
postgres=# delete from test_dt_int2;
DELETE 1
postgres=# INSERT INTO test_dt VALUES ('SELECT'), ('INSERT'), ('UPDATE'), ('DELETE');
INSERT 0 4
postgres=# INSERT INTO test_dt_int2 VALUES ('11,22');
INSERT 0 1
postgres=# select * from test_dt;
   c1   
--------
 SELECT
 INSERT
 UPDATE
 DELETE
(4 rows)

postgres=# select * from test_dt_int2;
  c1   
-------
 11,22
(1 row)

```

docs/09_datatypes.md
lyupan: 1. The most important reason is only `plain` can be used with fixed-length types;
2. In our previous implementation and the `CREATE TYPE` command, https://www.postgresql.org/docs/current/sql-createtype.html
```
storage
The storage strategy for the data type. If specified, must be plain, external, extended, or main; the default is plain.
```
the default storage is `plain`.

docs/09_datatypes.md
lyupan: See previous comment https://github.com/aws/pg_tle/pull/266#discussion_r1501067348.

src/datatype.c
lyupan: Thanks for the suggestion, updated!

src/datatype.c
lyupan: done

anth0nyleung (approved):

adamguo0 (approved):

docs/09_datatypes.md
jkatz: OK; this aligns with upstream.

src/datatype.c
```diff
@@ -55,6 +55,17 @@ static void check_pgtle_base_type(Oid typeOid);
...
+static char get_type_storage(char *storageStr);
+static Datum
+			pg_tle_create_base_type_internal(Oid typeNamespace,
+											 char *typeName,
+											 Oid inputFuncId,
+											 Oid outputFuncId,
+											 int16 internalLength,
+											 char *alignmentStr,
+											 char *storageStr,
+											 char *funcProbin);
```
jkatz, line 68: Was this the output from pgindent?

src/datatype.c
```diff
@@ -159,6 +170,57 @@ pg_tle_create_shell_type_if_not_exists(PG_FUNCTION_ARGS)
...
+	char	   *funcProbin = get_probin(fcinfo->flinfo->fn_oid);
+
+	return pg_tle_create_base_type_internal(typeNamespace, typeName, inputFuncId, outputFuncId, internalLength, alignment, storage, funcProbin);
+}
+
+/*
+ * pg_tle_create_base_type_with_storage
+ *
+ * Similar to pg_tle_create_base_type, but 7 args are expected.
+ * Two additional args are alignment and storage.
```
jkatz, line 190:
```suggestion
 * Similar to pg_tle_create_base_type, but expects additional arguments
 * for alignment and storage.
```

jkatz (approved):
LGTM, a couple of minor comments. Nice job!

src/datatype.c
lyupan: Yes, it is.

lyupan:
Thank you all for the review!

Issue #265, opened 2024-01-30
Clientauth skip lists do not work when user/database names are mixed case


adamguo0:
## Description

Users/databases on `pgtle.clientauth_users_to_skip` and `pgtle.clientauth_databases_to_skip` do not take effect if their names are mixed case.

## Steps to reproduce

1. Register a clientauth function
2. Create a role with a mixed case name, e.g. `CREATE ROLE "MixedCase";`
3. Add it to `pgtle.clientauth_users_to_skip` and reload config: `ALTER SYSTEM SET pgtle.clientauth_users_to_skip = 'MixedCase'; SELECT pg_reload_conf()`
4. Connect as `MixedCase`

## Expected outcome

Clientauth function should be skipped for `MixedCase`.

## Actual outcome

Clientauth function is executed.

This behaviour can be reproduced similarly for mixed-case database names.

TianzeMYou:
Merged

PR #268, opened 2024-02-29
Allow mixed case in pgtle.clientauth_users_to_skip and pgtle.clientauth_databases_to_skip

TianzeMYou:
…th_databases_to_skip

Issue #, if available: [265](https://github.com/aws/pg_tle/issues/265)

Description of changes: Allow mixed case in pgtle.clientauth_users_to_skip and pgtle.clientauth_databases_to_skip

### DIFF
--- a/test/t/004_pg_tle_clientauth.pl
+++ b/test/t/004_pg_tle_clientauth.pl
@@ -255,6 +255,46 @@ $node->psql('postgres', q[
         END
     $$ LANGUAGE plpgsql], on_error_die => 1);
 $node->psql('postgres', qq[SELECT pgtle.register_feature('reject_testuser', 'clientauth')], on_error_die => 1);
+### 16. Allow mixedCase in pgtle.clientauth_users_to_skip
+$node->psql('postgres', 'CREATE ROLE \"testUser3\" LOGIN', stderr => \$psql_err);
+$node->psql('postgres', q[
+    CREATE FUNCTION reject_testUser3(port pgtle.clientauth_port_subset, status integer) RETURNS text AS $$
+        BEGIN
+            IF port.user_name = 'testUser3' THEN
+                RETURN 'testUser3 is not allowed to connect';
+            ELSE
+                RETURN '';
+            END IF;
+        END
+    $$ LANGUAGE plpgsql;]);
+$node->psql('postgres', qq[SELECT pgtle.register_feature('reject_testUser3', 'clientauth')]);
+$node->psql('postgres', 'select', extra_params => ['-U', 'testUser3'], stderr => \$psql_err);
+like($psql_err, qr/FATAL:  testUser3 is not allowed to connect/,
+    "clientauth function rejects testUser3");
+
+$node->append_conf('postgresql.conf', qq(pgtle.enable_clientauth = 'on'));
+$node->append_conf('postgresql.conf', qq(pgtle.clientauth_users_to_skip = 'testUser3'));
+$node->restart;
+
+$node->command_ok(
+    ['psql', '-U', 'testUser3', '-c', 'select;'],
+    "clientauth function does not reject testUser3 when testUser3 is in pgtle.clientauth_users_to_skip");
+### 17. Allow mixedCase in pgtle.clientauth_databases_to_skip
+$node->psql('postgres', 'CREATE DATABASE \"mixedCaseDb\"');
+$node->append_conf('postgresql.conf', qq(pgtle.clientauth_users_to_skip = ''));
+$node->append_conf('postgresql.conf', qq(pgtle.clientauth_databases_to_skip = ''));
+$node->psql('postgres', 'SELECT pg_reload_conf();');
+
+$node->psql('mixedCaseDb', 'select', extra_params => ['-U', 'testUser3'], stderr => \$psql_err);
+like($psql_err, qr/FATAL:  testUser3 is not allowed to connect/,
+    "clientauth function rejects testUser3");
+
+$node->append_conf('postgresql.conf', qq(pgtle.clientauth_databases_to_skip = 'mixedCaseDb'));
+$node->psql('postgres', 'SELECT pg_reload_conf();');
+
+$node->command_ok(
+    ['psql', '-U', 'testUser3', '-c', 'select;'],
+    "clientauth function does not reject testUser3 when database is in pgtle.clientauth_databases_to_skip");
 # Create role with name "
 $node->psql('postgres', qq[CREATE ROLE """" LOGIN], on_error_die => 1);
 # Create role with name ""
--- a/src/feature.c
+++ b/src/feature.c
@@ -137,7 +137,7 @@ check_string_in_guc_list(const char *str, const char *guc_var, const char *guc_n
 	ListCell   *lc;
 
 	guc_copy = pstrdup(guc_var);
-	if (!SplitIdentifierString(guc_copy, ',', &guc_list))
+	if (!SplitGUCList(guc_copy, ',', &guc_list))
 		elog(ERROR, "could not parse %s", guc_name);
 
 	foreach(lc, guc_list)


test/t/004_pg_tle_clientauth.pl
```diff
@@ -255,6 +255,46 @@
         END
     $$ LANGUAGE plpgsql], on_error_die => 1);
 $node->psql('postgres', qq[SELECT pgtle.register_feature('reject_testuser', 'clientauth')], on_error_die => 1);
+### 16. Allow mixedCase in pgtle.clientauth_users_to_skip
+$node->psql('postgres', 'CREATE ROLE \"testUser3\" LOGIN', stderr => \$psql_err);
```
adamguo0, line 259: I think this line is failing because the double-quotes do not need to be escaped, i.e. `CREATE ROLE "testUser3"`.

Can you also add `on_error_die => 1` to each psql command that is not supposed to error? By default psql errors don't cause tap tests to fail which can be confusing.

test/t/004_pg_tle_clientauth.pl
```diff
@@ -255,6 +255,46 @@
...
+
+$node->append_conf('postgresql.conf', qq(pgtle.enable_clientauth = 'on'));
+$node->append_conf('postgresql.conf', qq(pgtle.clientauth_users_to_skip = 'testUser3'));
+$node->restart;
+
+$node->command_ok(
+    ['psql', '-U', 'testUser3', '-c', 'select;'],
+    "clientauth function does not reject testUser3 when testUser3 is in pgtle.clientauth_users_to_skip");
+### 17. Allow mixedCase in pgtle.clientauth_databases_to_skip
+$node->psql('postgres', 'CREATE DATABASE \"mixedCaseDb\"');
```
adamguo0, line 283: No need to escape quotes

test/t/004_pg_tle_clientauth.pl
```diff
@@ -255,6 +255,46 @@
...
+
+$node->psql('mixedCaseDb', 'select', extra_params => ['-U', 'testUser3'], stderr => \$psql_err);
+like($psql_err, qr/FATAL:  testUser3 is not allowed to connect/,
+    "clientauth function rejects testUser3");
+
+$node->append_conf('postgresql.conf', qq(pgtle.clientauth_databases_to_skip = 'mixedCaseDb'));
+$node->psql('postgres', 'SELECT pg_reload_conf();');
+
+$node->command_ok(
+    ['psql', '-U', 'testUser3', '-c', 'select;'],
```
adamguo0, line 296: This should specify `-d mixedCaseDb`

anth0nyleung:
Please also remove `…th_databases_to_skip` from the description of the pr

test/t/004_pg_tle_clientauth.pl
```diff
@@ -255,6 +255,46 @@
...
+        END
+    $$ LANGUAGE plpgsql;]);
+$node->psql('postgres', qq[SELECT pgtle.register_feature('reject_testUser3', 'clientauth')]);
+$node->psql('postgres', 'select', extra_params => ['-U', 'testUser3'], stderr => \$psql_err);
+like($psql_err, qr/FATAL:  testUser3 is not allowed to connect/,
+    "clientauth function rejects testUser3");
+
+$node->append_conf('postgresql.conf', qq(pgtle.enable_clientauth = 'on'));
+$node->append_conf('postgresql.conf', qq(pgtle.clientauth_users_to_skip = 'testUser3'));
+$node->restart;
```
anth0nyleung, line 277: why do we need to restart?

test/t/004_pg_tle_clientauth.pl
```diff
@@ -255,6 +255,34 @@
         END
     $$ LANGUAGE plpgsql], on_error_die => 1);
 $node->psql('postgres', qq[SELECT pgtle.register_feature('reject_testuser', 'clientauth')], on_error_die => 1);
+### 16. Allow mixedCase in pgtle.clientauth_users_to_skip
```
adamguo0, line 258: Can you move your tests to after the line
```
    "role with injection payload in name can connect");
```
That block is supposed to be part of `15. Malformed strings cannot be used for SQL injection`

test/t/004_pg_tle_clientauth.pl
```diff
@@ -256,45 +256,33 @@
...
             IF port.user_name = 'testUser3' THEN
                 RETURN 'testUser3 is not allowed to connect';
-            ELSE
-                RETURN '';
             END IF;
         END
     $$ LANGUAGE plpgsql;]);
-$node->psql('postgres', qq[SELECT pgtle.register_feature('reject_testUser3', 'clientauth')]);
-$node->psql('postgres', 'select', extra_params => ['-U', 'testUser3'], stderr => \$psql_err);
-like($psql_err, qr/FATAL:  testUser3 is not allowed to connect/,
```
adamguo0: I'd prefer to keep this test here so we know that clientauth was set up properly to reject testUser3 before adding it to skip list

test/t/004_pg_tle_clientauth.pl
```diff
@@ -256,45 +256,33 @@
...
     "clientauth function does not reject testUser3 when testUser3 is in pgtle.clientauth_users_to_skip");
 ### 17. Allow mixedCase in pgtle.clientauth_databases_to_skip
-$node->psql('postgres', 'CREATE DATABASE \"mixedCaseDb\"');
+$node->psql('postgres', 'CREATE DATABASE "mixedCaseDb"');
 $node->append_conf('postgresql.conf', qq(pgtle.clientauth_users_to_skip = ''));
-$node->append_conf('postgresql.conf', qq(pgtle.clientauth_databases_to_skip = ''));
-$node->psql('postgres', 'SELECT pg_reload_conf();');
-
-$node->psql('mixedCaseDb', 'select', extra_params => ['-U', 'testUser3'], stderr => \$psql_err);
-like($psql_err, qr/FATAL:  testUser3 is not allowed to connect/,
```
adamguo0: Same here

test/t/004_pg_tle_clientauth.pl
TianzeMYou: I digged into this a bit and I think `stderr => \$psql_err` is a bit better because `on_error_die => 1` just fails the test but doesn't offer any output or tell us which line the test fails on

adamguo0: In that case you'll need to verify that `psql_err` is empty after each `psql` invocation, otherwise it will populate `psql_err` with stderr but won't throw an error

adamguo0 (approved):

anth0nyleung:
It's not very clear to me why we need to move the new tests to a separate file.

test/t/004_pg_tle_clientauth.pl
anth0nyleung: Cant we have both stderr and on_error_die as arguments in the psql call?

adamguo0: If we specify on_error_die and psql throws an error, it won't print psql_err by default iiuc. We still need a line such as `like($psql_err, '')`

adamguo0: `on_error_die` also emits the psql error to the regress logs. They should be uploaded as artifacts if a test run fails

adamguo0:
Do you mind adding your test case and updating the numbers at the top of the file as well: https://github.com/aws/pg_tle/blob/main/test/t/004_pg_tle_clientauth.pl#L24

anth0nyleung:
> Do you mind adding your test case and updating the numbers at the top of the file as well: https://github.com/aws/pg_tle/blob/main/test/t/004_pg_tle_clientauth.pl#L24

+1 on this

adamguo0:
LGTM

anth0nyleung (approved):

Issue #269, opened 2024-03-12
Warn the user if a hook function is registered but a necessary parameter is not enabled


adamguo0:
## Describe the problem

Currently a user can register a hook function regardless of whether the necessary parameters are enabled. For instance, a user can register a clientauth function without setting `pgtle.enable_clientauth` or `pgtle.clientauth_db_name`, both of which are needed to enable the clientauth feature. This can be confusing for users who are not familiar with the feature.

## Describe the proposal

Emit a WARNING when registering a hook function if the parameters aren't enabled.

- Passcheck: warn if `pgtle.enable_password_check` is OFF or if `pgtle.passcheck_db_name` is not the current database
- Clientauth: warn if `pgtle.enable_clientauth` is OFF or if `pgtle.clientauth_db_name` is not the current database

These warnings will help to preempt configuration errors.

Relevant code: https://github.com/aws/pg_tle/blob/main/pg_tle--1.1.1.sql#L601-L630

## Describe alternatives

Currently a user can read the documentation https://github.com/aws/pg_tle/blob/main/docs/04_hooks.md to debug their configuration. However we should still try to make the user experience smoother with some helpful hints.

PR #270, opened 2024-03-14
Warn the user if a hook function is registered but a necessary parameter is not enabled

TianzeMYou:
Issue #, if available: https://github.com/aws/pg_tle/issues/269

Description of changes: Warn the user if a hook function is registered but a necessary parameter is not enabled

### DIFF
--- a/pg_tle--1.3.4--1.4.0.sql
+++ b/pg_tle--1.3.4--1.4.0.sql
@@ -78,6 +78,7 @@ clientauth_enabled text;
 current_db text;
 passcheck_db text;
 clientauth_db text;
+notice_message text;
 
 BEGIN
     SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.enable_password_check' INTO passcheck_enabled;
@@ -88,23 +89,23 @@ BEGIN
 
     IF feature = 'passcheck' THEN
         IF passcheck_enabled = 'off' THEN
-           RAISE NOTICE 'pgtle.enable_password_check is set to "off". To enable passcheck, set pgtle.enable_password_check = "on"';
+           RAISE NOTICE 'pgtle.enable_password_check is set to off. To enable passcheck, set pgtle.enable_password_check = on';
         ELSE
         -- passcheck_db_name is an optional param, we only emit a warning if it's non-empty and is not the current database
             IF passcheck_db != '' AND current_db != passcheck_db THEN
-                RAISE NOTICE 'pgtle.passcheck_db_name is currently "%". To trigger this passcheck function, register the function in that database.', passcheck_db
-                USING HINT = pg_catalog.FORMAT('Alternatively, to use the current database for passcheck, set pgtle.passcheck_db_name = "%s" and reload the PostgreSQL configuration.', current_db);
+                RAISE NOTICE '%', pg_catalog.FORMAT('pgtle.passcheck_db_name is currently %I. To trigger this passcheck function, register the function in that database.', passcheck_db)
+                USING HINT = pg_catalog.FORMAT('Alternatively, to use the current database for passcheck, set pgtle.passcheck_db_name = %I and reload the PostgreSQL configuration.', current_db);
             END IF;
         END IF;
     END IF;
 
     IF feature = 'clientauth' THEN
         IF clientauth_enabled = 'off' THEN
-            RAISE NOTICE 'pgtle.enable_clientauth is set to "off". To enable clientauth, set pgtle.enable_clientauth = "on"';
+            RAISE NOTICE 'pgtle.enable_clientauth is set to off. To enable clientauth, set pgtle.enable_clientauth = on';
         ELSE
             IF current_db != clientauth_db THEN
-                RAISE NOTICE 'pgtle.clientauth_db_name is currently "%". To trigger this clientauth function, register the function in that database.', clientauth_db
-                USING HINT = pg_catalog.FORMAT('Alternatively, to use the current database for clientauth, set pgtle.clientauth_db_name = "%s" and reload the PostgreSQL configuration.', current_db);
+                RAISE NOTICE '%', pg_catalog.FORMAT('pgtle.clientauth_db_name is currently %I. To trigger this clientauth function, register the function in that database.', clientauth_db)
+                USING HINT = pg_catalog.FORMAT('Alternatively, to use the current database for clientauth, set pgtle.clientauth_db_name = %I and reload the PostgreSQL configuration.', current_db);
             END IF;
         END IF;
     END IF;
--- a/pg_tle--1.3.4--1.4.0.sql
+++ b/pg_tle--1.3.4--1.4.0.sql
@@ -88,11 +88,11 @@ BEGIN
 
     IF feature = 'passcheck' THEN
         IF passcheck_enabled = 'off' THEN
-           RAISE NOTICE 'Required parameter pgtle.enable_password_check is "off". To enable passcheck, set pgtle.enable_password_check = "on"';
+           RAISE NOTICE 'pgtle.enable_password_check is set to "off". To enable passcheck, set pgtle.enable_password_check = "on"';
         ELSE
         -- passcheck_db_name is an optional param, we only emit a warning if it's non-empty and is not the current database
             IF passcheck_db != '' AND current_db != passcheck_db THEN
-                RAISE NOTICE 'pgtle.passcheck_db_name is currently %. To trigger this passcheck function, register the function in that database.', passcheck_db
+                RAISE NOTICE 'pgtle.passcheck_db_name is currently "%". To trigger this passcheck function, register the function in that database.', passcheck_db
                 USING HINT = pg_catalog.FORMAT('Alternatively, to use the current database for passcheck, set pgtle.passcheck_db_name = "%s" and reload the PostgreSQL configuration.', current_db);
             END IF;
         END IF;
@@ -100,10 +100,10 @@ BEGIN
 
     IF feature = 'clientauth' THEN
         IF clientauth_enabled = 'off' THEN
-            RAISE NOTICE 'Required parameter pgtle.enable_clientauth is "off". To enable clientauth, set pgtle.enable_clientauth = "on"';
+            RAISE NOTICE 'pgtle.enable_clientauth is set to "off". To enable clientauth, set pgtle.enable_clientauth = "on"';
         ELSE
             IF current_db != clientauth_db THEN
-                RAISE NOTICE 'pgtle.clientauth_db_name is currently %. To trigger this clientauth function, register the function in that database.', clientauth_db
+                RAISE NOTICE 'pgtle.clientauth_db_name is currently "%". To trigger this clientauth function, register the function in that database.', clientauth_db
                 USING HINT = pg_catalog.FORMAT('Alternatively, to use the current database for clientauth, set pgtle.clientauth_db_name = "%s" and reload the PostgreSQL configuration.', current_db);
             END IF;
         END IF;
--- a/pg_tle--1.3.4--1.4.0.sql
+++ b/pg_tle--1.3.4--1.4.0.sql
@@ -19,7 +19,6 @@
 
 DROP FUNCTION pgtle.create_base_type CASCADE;
 DROP FUNCTION pgtle.create_base_type_if_not_exists CASCADE;
--- DROP FUNCTION pgtle.register_feature CASCADE;
 
 CREATE FUNCTION pgtle.create_base_type
 (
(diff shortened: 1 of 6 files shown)


pg_tle--1.3.4--1.4.0.sql
```diff
@@ -60,6 +61,71 @@ END;
...
+BEGIN
+    SELECT setting FROM pg_settings WHERE name = 'pgtle.enable_password_check' INTO passcheck_enabled;
+    SELECT setting FROM pg_settings WHERE name = 'pgtle.enable_clientauth' INTO clientauth_enabled;
+    SELECT CURRENT_DATABASE() INTO current_db;
+    SELECT setting FROM pg_settings WHERE name = 'pgtle.passcheck_db_name' INTO passcheck_db;
+    SELECT setting FROM pg_settings WHERE name = 'pgtle.clientauth_db_name' INTO clientauth_db;
+
+    IF feature = 'passcheck' THEN
+        IF passcheck_enabled = 'off' THEN
+           RAISE WARNING 'Required parameter pgtle.enable_password_check is off. To enable passcheck, run ALTER SYSTEM SET pgtle.enable_password_check = "on"';
```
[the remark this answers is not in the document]
TianzeMYou: hmm I think warning is okay since we do _need_ to have parameters set to 'on' for user hooks to work. open to suggestions though.

pg_tle--1.3.4--1.4.0.sql
```diff
@@ -60,6 +61,71 @@ END;
...
+proc_schema_name text;
+ident text;
+passcheck_enabled text;
+clientauth_enabled text;
+current_db text;
+passcheck_db text;
+clientauth_db text;
+
+BEGIN
+    SELECT setting FROM pg_settings WHERE name = 'pgtle.enable_password_check' INTO passcheck_enabled;
```
anth0nyleung, line 84: let's schema qualify them - add pg_catalog in front of pg_settings and CURRENT_DATABASE

pg_tle--1.3.4--1.4.0.sql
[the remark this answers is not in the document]
anth0nyleung: I dont think we should say `run ALTER SYSTEM SET pgtle.enable_password_check = "on"'`, instead i prefer to just mention that this parameter needs to be enabled, since we do not allow customers doing ALTER SYSTEM SET in RDS, but in open source it's up to admin to do that so it could be confusing.

IMO, warning usually means something unexpected has happened internally, e.g. https://github.com/postgres/postgres/blob/84c18acaf690e438e953e387caf1c13298d4ecb4/src/test/regress/sql/brin.sql#L353

https://github.com/postgres/postgres/blob/84c18acaf690e438e953e387caf1c13298d4ecb4/src/test/regress/sql/brin.sql#L368

In this case, i dont think it qualifies as "something unexpected has happened in the internal logic", so i think we can lower it to NOTICE

test/sql/pg_tle_api.sql
```diff
@@ -230,6 +230,43 @@ SELECT EXISTS(
 );
 
 SELECT pgtle.uninstall_extension('test_unregister_feature');
+-- create this function one more time to test warning messages
+CREATE FUNCTION password_check_length_greater_than_8(username text, shadow_pass text, password_types pgtle.password_types, validuntil_time TimestampTz,validuntil_null boolean) RETURNS void AS
```
anth0nyleung, line 234: nit: missing space after `TimestampTz`

anth0nyleung:
I think we can ask @jkatz to take a look at the customer-facing wording / messages in this PR.

anth0nyleung:
Could you also test upgrading from 1.3 -> 1.4 to ensure that it succeeds

TianzeMYou:
`postgres=# create extension pg_tle version '1.3.4';      
CREATE EXTENSION
postgres=# CREATE FUNCTION password_check_length_greater_than_8(username text, shadow_pass text, password_types pgtle.password_types, validuntil_time TimestampTz, validuntil_null boolean) RETURNS void AS
    $$
postgres-# postgres$#     BEGIN
        if length(shadow_pass) < 8 THEN
postgres$# postgres$#             RAISE EXCEPTION 'Passwords needs to be longer than 8';
        END IF;
    END;
    $$
Lpostgres$# postgres$# postgres$# postgres-# ANGUAGE PLPGSQL;
CREATE FUNCTION
postgres=# CREATE FUNCTION password_check_length_greater_than_7(username text, shadow_pass text, password_types pgtle.password_types, validuntil_time TimestampTz, validuntil_null boolean) RETURNS void AS
    $$
postgres-# postgres$#     BEGIN
        if length(shadow_pass) < 8 THEN
postgres$# postgres$#             RAISE EXCEPTION 'Passwords needs to be longer than 8';
        END IF;
    END;
    $$
Lpostgres$# postgres$# postgres$# postgres-# ANGUAGE PLPGSQL;
CREATE FUNCTION
postgres=# alter extension pg_tle update 1.4.0
postgres-# ;
ERROR:  syntax error at or near "1.4"
LINE 1: alter extension pg_tle update 1.4.0
                                      ^
postgres=# alter extension pg_tle '1.4.0';
ERROR:  syntax error at or near "'1.4.0'"
LINE 1: alter extension pg_tle '1.4.0';
                               ^
postgres=# alter extension pg_tle update 1.4.0
postgres-# ;
ERROR:  syntax error at or near "1.4"
LINE 1: alter extension pg_tle update 1.4.0
                                      ^
postgres=# ^^C
postgres=# ^C
postgres=# alter extension pg_tle update to 1.4.0
postgres-# ;
ERROR:  syntax error at or near "1.4"
LINE 1: alter extension pg_tle update to 1.4.0
                                         ^
postgres=# alter extension pg_tle update to '1.4.0';
ALTER EXTENSION
postgres=# SELECT pgtle.register_feature('password_check_length_greater_than_8', 'passcheck');
NOTICE:  Required parameter pgtle.enable_password_check is "off". To enable passcheck, set pgtle.enable_password_check = "on"
 register_feature 
------------------
 
(1 row)

postgres=# alter system set pgtle.enable_password_check = "on";
ALTER SYSTEM
postgres=# SELECT pgtle.register_feature('password_check_length_greater_than_8', 'passcheck');^[[D^[[D^C  
postgres=# SELECT pgtle.register_feature('password_check_length_greater_than_7', 'passcheck'); 
NOTICE:  Required parameter pgtle.enable_password_check is "off". To enable passcheck, set pgtle.enable_password_check = "on"
 register_feature 
------------------
 
(1 row)

postgres=# \c -
You are now connected to database "postgres" as user "michaeey".
postgres=# SELECT pgtle.register_feature('password_check_length_greater_than_7', 'passcheck');
NOTICE:  Required parameter pgtle.enable_password_check is "off". To enable passcheck, set pgtle.enable_password_check = "on"
ERROR:  duplicate key value violates unique constraint "feature_info_pkey"
DETAIL:  Key (feature, schema_name, proname)=(passcheck, public, password_check_length_greater_than_7) already exists.
CONTEXT:  SQL statement "INSERT INTO pgtle.feature_info VALUES (feature, proc_schema_name, proname, ident)"
PL/pgSQL function pgtle.register_feature(regproc,pgtle.pg_tle_features) line 58 at SQL statement
postgres=# drop function pgtle.register_feature('password_check_length_greater_than_7', 'passcheck');
ERROR:  syntax error at or near "'password_check_length_greater_than_7'"
LINE 1: drop function pgtle.register_feature('password_check_length_...
                                             ^
postgres=# drop function password_check_length_greater_than_7;
ERROR:  Function is referenced in pgtle.feature_info
CONTEXT:  PL/pgSQL function pgtle.pg_tle_feature_info_sql_drop() line 27 at RAISE
postgres=# drop extension pg_tle cascade;
NOTICE:  drop cascades to 2 other objects
DETAIL:  drop cascades to function password_check_length_greater_than_8(text,text,pgtle.password_types,timestamp with time zone,boolean)
drop cascades to function password_check_length_greater_than_7(text,text,pgtle.password_types,timestamp with time zone,boolean)
DROP EXTENSION
postgres=# create extension pg_tle version '1.3.4';
CREATE EXTENSION
postgres=# CREATE FUNCTION testhook1(username text, shadow_pass text, password_types pgtle.password_types, validuntil_time TimestampTz, validuntil_null boolean) RETURNS void AS
    $$
    BEGIN
        if lengthpostgres-# postgres$# postgres$# (shadow_pass) < 8 THEN
            RAISE EXCEPTIONpostgres$#  'Passwords needs to be longer than 8';
        ENpostgres$# D IF;
    END;
    $$
LANGUAGE PLPGSQL;postgres$# postgres$# postgres-# 
CREATE FUNCTION
postgres=# CREATE FUNCTION testhook2(username text, shadow_pass text, password_types pgtle.password_types, validuntil_time TimestampTz, validuntil_null boolean) RETURNS void AS
    $$
    BEGIN
        if lengthpostgres-# postgres$# postgres$# (shadow_pass) < 7 THEN
            RAISE EXCEPTIONpostgres$#  'Passwords needs to be longer than 8';
        ENpostgres$# D IF;
    END;
    $$
LANGUAGE PLPGSQL;postgres$# postgres$# postgres-# 
CREATE FUNCTION
postgres=# ALTER SYSTEM SET pgtle.enable_password_check = 'off';
ALTER SYSTEM
postgres=# SELECT pg_reload_conf();
 pg_reload_conf 
----------------
 t
(1 row)

postgres=# SELECT pgtle.register_feature('testhook1', 'passcheck');
 register_feature 
------------------
 
(1 row)

postgres=# alter extension pg_tle update to '1.4.0';
ALTER EXTENSION
postgres=# SELECT pgtle.register_feature('testhook2', 'passcheck');
NOTICE:  Required parameter pgtle.enable_password_check is "off". To enable passcheck, set pgtle.enable_password_check = "on"
 register_feature 
------------------
 
(1 row)

postgres=# alter system set pgtle.enable_password_check = "on";
ALTER SYSTEM
postgres=# SELECT pg_reload_conf();
 pg_reload_conf 
----------------
 t
(1 row)

postgres=# SELECT pgtle.unregister_feature('testhook2', 'passcheck');
 unregister_feature 
--------------------
 
(1 row)

postgres=# SELECT pgtle.register_feature('testhook2', 'passcheck');
NOTICE:  pgtle.passcheck_db_name is currently contrib_regression. To trigger this passcheck function, register the function in that database.
HINT:  Alternatively, to use the current database for passcheck, set pgtle.passcheck_db_name = "postgres" and reload the PostgreSQL configuration.
 register_feature 
------------------

...
 
(1 row)`

pg_tle--1.3.4--1.4.0.sql
```diff
@@ -60,6 +60,71 @@ END;
...
+BEGIN
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.enable_password_check' INTO passcheck_enabled;
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.enable_clientauth' INTO clientauth_enabled;
+    SELECT pg_catalog.CURRENT_DATABASE() INTO current_db;
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.passcheck_db_name' INTO passcheck_db;
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.clientauth_db_name' INTO clientauth_db;
+
+    IF feature = 'passcheck' THEN
+        IF passcheck_enabled = 'off' THEN
+           RAISE NOTICE 'Required parameter pgtle.enable_password_check is "off". To enable passcheck, set pgtle.enable_password_check = "on"';
```
adamguo0, line 91: nit: imo it would be better to remove the word `required`, it's only "required" if you're enabling the feature right away. If you're just doing setup then this is a valid thing to do. Open to other suggestions

`Required parameter -> The parameter`?

pg_tle--1.3.4--1.4.0.sql
```diff
@@ -60,6 +60,71 @@ END;
...
+            IF passcheck_db != '' AND current_db != passcheck_db THEN
+                RAISE NOTICE 'pgtle.passcheck_db_name is currently %. To trigger this passcheck function, register the function in that database.', passcheck_db
+                USING HINT = pg_catalog.FORMAT('Alternatively, to use the current database for passcheck, set pgtle.passcheck_db_name = "%s" and reload the PostgreSQL configuration.', current_db);
+            END IF;
+        END IF;
+    END IF;
+
+    IF feature = 'clientauth' THEN
+        IF clientauth_enabled = 'off' THEN
+            RAISE NOTICE 'Required parameter pgtle.enable_clientauth is "off". To enable clientauth, set pgtle.enable_clientauth = "on"';
```
adamguo0, line 103: nit: same as above

pg_tle--1.3.4--1.4.0.sql
```diff
@@ -60,6 +60,71 @@ END;
...
+current_db text;
+passcheck_db text;
+clientauth_db text;
+
+BEGIN
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.enable_password_check' INTO passcheck_enabled;
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.enable_clientauth' INTO clientauth_enabled;
+    SELECT pg_catalog.CURRENT_DATABASE() INTO current_db;
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.passcheck_db_name' INTO passcheck_db;
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.clientauth_db_name' INTO clientauth_db;
```
jkatz, line 87: Do we need any `NULL` handling in the select list?

pg_tle--1.3.4--1.4.0.sql
jkatz, line 91:
```suggestion
           RAISE NOTICE 'pgtle.enable_password_check is set to "off". To enable passcheck, set pgtle.enable_password_check = "on"';
```

jkatz: Is there a reason this is a notice and not an error?

pg_tle--1.3.4--1.4.0.sql
```diff
@@ -60,6 +60,71 @@ END;
 $_pgtleie_$
 LANGUAGE plpgsql STRICT;
 
+-- Helper function to register features in the feature_info table
+CREATE OR REPLACE FUNCTION pgtle.register_feature(proc regproc, feature pgtle.pg_tle_features)
```
jkatz, line 64: I'm curious why we're calling this function `register_feature` after the generic registration function? I thought the idea of the of the original `register_feature` function is that it'd be agnostic to the different features (hooks) available, so we could have a single interface that enables them.

pg_tle--1.3.4--1.4.0.sql
adamguo0: This is still a valid usecase, since the user may be registering functions without intending to use them yet. However we got some feedback that it may be confusing if users register their functions and don't know why they're not triggering, especially since there are multiple parameters. So we just want to emit a notice in case they're wondering.

pg_tle--1.3.4--1.4.0.sql
adamguo0: Functionality-wise this is the same function. The only non-feature-agnostic part is where we emit NOTICES based on the param values.

i.e. if we add a new feature we don't need to change this function definition, unless we want to add another notice message

pg_tle--1.3.4--1.4.0.sql
TianzeMYou: I believe we do not. Testing should already cover scenarios where some value is not set.

pg_tle--1.3.4--1.4.0.sql
TianzeMYou: I think the `required` still makes sense here because unless that parameter is "on", the hook will not function.
i.e. it's `required` if user want the hook they are registering to work. It is not optional like the db_name.

pg_tle--1.3.4--1.4.0.sql
```diff
@@ -60,6 +60,71 @@ END;
...
+            END IF;
+        END IF;
+    END IF;
+
+    IF feature = 'clientauth' THEN
+        IF clientauth_enabled = 'off' THEN
+            RAISE NOTICE 'Required parameter pgtle.enable_clientauth is "off". To enable clientauth, set pgtle.enable_clientauth = "on"';
+        ELSE
+            IF current_db != clientauth_db THEN
+                RAISE NOTICE 'pgtle.clientauth_db_name is currently %. To trigger this clientauth function, register the function in that database.', clientauth_db
```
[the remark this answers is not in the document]
TianzeMYou: thanks, resolved

pg_tle--1.3.4--1.4.0.sql
```diff
@@ -60,6 +60,71 @@ END;
...
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.passcheck_db_name' INTO passcheck_db;
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.clientauth_db_name' INTO clientauth_db;
+
+    IF feature = 'passcheck' THEN
+        IF passcheck_enabled = 'off' THEN
+           RAISE NOTICE 'Required parameter pgtle.enable_password_check is "off". To enable passcheck, set pgtle.enable_password_check = "on"';
+        ELSE
+        -- passcheck_db_name is an optional param, we only emit a warning if it's non-empty and is not the current database
+            IF passcheck_db != '' AND current_db != passcheck_db THEN
+                RAISE NOTICE 'pgtle.passcheck_db_name is currently %. To trigger this passcheck function, register the function in that database.', passcheck_db
```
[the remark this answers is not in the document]
TianzeMYou: resolved

adamguo0 (approved):

pg_tle--1.3.4--1.4.0.sql
jkatz: I think we may want to rethink how we're able to get in custom notices on a per-feature basis, because this is likely to happen again. However, that appears to be a two-way door based on how this is organized

pg_tle--1.3.4--1.4.0.sql
```diff
@@ -60,6 +60,71 @@ END;
...
+    SELECT setting FROM pg_catalog.pg_settings WHERE name = 'pgtle.clientauth_db_name' INTO clientauth_db;
+
+    IF feature = 'passcheck' THEN
+        IF passcheck_enabled = 'off' THEN
+           RAISE NOTICE 'pgtle.enable_password_check is set to "off". To enable passcheck, set pgtle.enable_password_check = "on"';
+        ELSE
+        -- passcheck_db_name is an optional param, we only emit a warning if it's non-empty and is not the current database
+            IF passcheck_db != '' AND current_db != passcheck_db THEN
+                RAISE NOTICE 'pgtle.passcheck_db_name is currently "%". To trigger this passcheck function, register the function in that database.', passcheck_db
+                USING HINT = pg_catalog.FORMAT('Alternatively, to use the current database for passcheck, set pgtle.passcheck_db_name = "%s" and reload the PostgreSQL configuration.', current_db);
```
jkatz, line 96: Consider `%I` instead of `%s` for SQL identifier. That would also handle the quoting.

TianzeMYou: `RAISE NOTICE 'pgtle.enable_password_check is set to "off". To enable passcheck, set pgtle.enable_password_check = "on"';`
this line always has "" around off or on
```
RAISE NOTICE '%', pg_catalog.FORMAT('pgtle.passcheck_db_name is currently %I. To trigger this passcheck function, register the function in that database.', passcheck_db)
USING HINT = pg_catalog.FORMAT('Alternatively, to use the current database for passcheck, set pgtle.passcheck_db_name = %I and reload the PostgreSQL configuration.', current_db);
```
if we use `%I`, lowercaseidentiers won't have "", while "mixedCaseIdentifiers"  would. Is that inconsistency ok? 

I think it's a trivial change, we should just have "" at all times since double quotes in psql classifies something as identifier and it is consistent

Another approach is just to take out "" from
`RAISE NOTICE 'pgtle.enable_password_check is set to "off". To enable passcheck, set pgtle.enable_password_check = "on"';`

That way all lowercasedstuff won't have "", making it consistent.

PR #272, opened 2024-04-05
Refactor typo in 04_hooks.md doc

Rajanpandey:
Update the SQL command to set enable_clientauth hook in the 2nd example in the hook doc.
It has a typo suggesting to activate enable_password_check hook instead.

### DIFF
--- a/docs/04_hooks.md
+++ b/docs/04_hooks.md
@@ -345,7 +345,7 @@ $_pgtle_$
 To enable the `clientauth` hook, you will need to set `pgtle.enable_clientauth` to `on` or `require` and restart the database. For example:
 
 ```sql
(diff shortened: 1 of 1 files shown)


adamguo0:
Good catch! One comment, `SELECT pg_catalog.pg_reload_conf()` is not sufficient to enable the `clientauth` feature, instead Postgres needs to be restarted. This is covered in line 352 so I think we can remove the `SELECT pg_catalog.pg_reload_conf()` line.

Rajanpandey:
Thanks @adamguo0 ! :D Made the changes.
Seems like I got multiple commits, I will do a squash merge so that only 1 commit is shown. (Haven't used github since a while, will figure out my way soon)! :)

Issue #274, opened 2024-04-18
Login attempts could repeat depending on SSLMode and leads to incorrect count of attempts


TianzeMYou:
## Description

For the lockout-after-a-specific-number-of-login-attempts example, login attempts could increment by 2 depending on SSLMode

https://www.postgresql.org/message-id/flat/17032-7ad69968e744de35%40postgresql.org#:~:text=It%20tries%20to%20connect%20with%20encryption%20and%20fails%2C%20then%20tries%20to%20connect%0Awithout%20encryption%20and%20also%20fails

## Steps to reproduce

Run the example here with sslmode set to allow/prefer
https://github.com/aws/pg_tle/blob/main/examples/client_lockout/client_lockout--1.0.sql

## Expected outcome

Client should be locked out after 5 attempts

## Actual outcome

Client is locked out before 5 attempts

## Analysis

Could potentially check the SSLMode in hook sql, which requires a C code change.
https://github.com/aws/pg_tle/blob/main/src/clientauth.c#L144-L166

adamguo0:
Since this is an expected and documented client-side behaviour [1], we shouldn't try to change this on the pg_tle side. However it's worth adding a note in the example for users to be aware.

[1] https://www.postgresql.org/docs/devel/libpq-connect.html#LIBPQ-CONNECT-SSLMODE

PR #275, opened 2024-04-22
Add note about libpq sslmode to clientauth docs

adamguo0:
Issue #, if available: #274

Description of changes: Add note about libpq sslmode to clientauth docs

### DIFF
--- a/docs/04_hooks.md
+++ b/docs/04_hooks.md
@@ -342,6 +342,8 @@ $_pgtle_$
 );
 ```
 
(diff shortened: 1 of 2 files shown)


TianzeMYou:
lgtm

docs/04_hooks.md
```diff
@@ -342,6 +342,8 @@ $_pgtle_$
 );
 ```
 
+> **Note**: If the database client has set the `sslmode` parameter to `allow` or `prefer`, the client will automatically attempt to re-connect if the first connection fails. This will trigger the `clientauth` function twice, adding 2 to their failed connection attempts. See the [PostgreSQL documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE) for more information.
```
anth0nyleung, line 345: Im thinking we might want to mention why this is important and what users might experience, like this behavior might cause the client to get locked out prematurely. Im okay with either adding it here or in the example README.md or both.

`This will trigger the `clientauth` function twice, adding two extra failed connection attempts, and potentially causing unexpected early lock out`

adamguo0: Yeah good point

Issue #126, opened 2022-10-18
`extension_update_paths(name name ...)` => `extension_update_paths(name text ...)`


jkatz:
All of the other extension functions use `text` as name, so this appears to be an arbitrary restriction.

Consider "relaxing" the length of the `name` input.

Issue #100, opened 2022-10-11
Misleading Error Message when extension name is > NAMEDATALEN


JohnHVancouver:
```
postgres=# SELECT pgtle.install_extension
(
 'test9greaterthanNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALEN',
 '0.1',
 'comment',
$_pg_tle_$                                                                         
    CREATE FUNCTION dist(x1 numeric, y1 numeric, x2 numeric, y2 numeric, l numeric)
    RETURNS numeric
    AS $$                                                                  
      SELECT ((x2 ^ l - x1 ^ l) ^ (1 / l)) + ((y2 ^ l - y1 ^ l) ^ (1 / l));
    $$ LANGUAGE SQL;
$_pg_tle_$
);
NOTICE:  identifier "test9greaterthanNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALEN.control" will be truncated to "test9greaterthanNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAM"
NOTICE:  identifier "test9greaterthanNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALEN--0.1.sql" will be truncated to "test9greaterthanNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAM"
ERROR:  Extension 'test9greaterthanNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALENNAMEDATALEN' already installed.
```

It says the extension's already installed even though this is our first time installing it.

PR #277, opened 2024-05-09
Relax extension_update_paths argument to text

Rajanpandey:
Issue #, if available: #126

Description of changes:
All of the other extension functions use text as name, so this appears to be an arbitrary restriction. Hence, relaxing the argument from name to text.

The `name` data (currently in `extension_update_paths()`) can hold 63 characters only (ref #define NAMEDATALEN 64). Whereas `text` data type (updated in this PR) has no limit at all.

Note that `install_extension()` already uses `text` instead of `name` data type for the extension-name's argument. So a user can create an extension with names that are >63 characters, but the operation will fail when trying to find its update paths using `extension_update_paths()` function because it's using a data type that limits the extension name to 63 chars!

### DIFF
--- a/src/tleextension.c
+++ b/src/tleextension.c
@@ -2808,7 +2808,7 @@ PG_FUNCTION_INFO_V1(pg_tle_extension_update_paths);
 Datum
 pg_tle_extension_update_paths(PG_FUNCTION_ARGS)
 {
-	Name		extname = PG_GETARG_NAME(0);
+	char*		extname = text_to_cstring(PG_GETARG_TEXT_P(0));
 	ReturnSetInfo *rsinfo = (ReturnSetInfo *) fcinfo->resultinfo;
 	List	   *evi_list;
 	ExtensionControlFile *control;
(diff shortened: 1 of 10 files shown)


src/tleextension.c
```diff
@@ -2808,7 +2808,7 @@ PG_FUNCTION_INFO_V1(pg_tle_extension_update_paths);
 Datum
 pg_tle_extension_update_paths(PG_FUNCTION_ARGS)
 {
-	Name		extname = PG_GETARG_NAME(0);
+	char*		extname = text_to_cstring(PG_GETARG_TEXT_P(0));
```
adamguo0, line 2811: To make pg_indent happy:
```suggestion
	char	   *extname = text_to_cstring(PG_GETARG_TEXT_P(0));
```

adamguo0:
To update the extension version you'll need to bump EXTVERSION in the Makefile as well. Not sure if that's causing these regression test failures though

Rajanpandey:
Thanks @adamguo0, fixed it! 😄

adamguo0:
Sorry I haven't gotten to this in a while -- the change looks good to me. I did a manual sanity check to make sure that after updating pg_tle from 1.4.0 to 1.4.1, the function works as expected

anth0nyleung:
Hi, do you mind elaborating what are the restrictions that you're facing before this change?

Rajanpandey:
Hi @anth0nyleung 😃, 
The name data type can hold 63 characters string only (ref `#define NAMEDATALEN 64`). Whereas text has no limit at all!

Note that `install_extension` already uses `text` instead of `name` data type for the extension-name's argument. So a user can create an extension with names that are >63 characters, but the operation will fail when trying to find its update paths using `extension_update_paths()` function because it's using a data type that limits the extension name to 63 chars!

This PR addresses a long time pending issue, #126 raised by @jkatz. 😄

anth0nyleung (approved):

adamguo0:
As a side-note there's another bug preventing users from creating extensions with almost-63 chars. We'll need to address that separately

https://github.com/aws/pg_tle/issues/100

Issue #284, opened 2025-01-23
How to specify extension schema?


mkleczek:
Hi, I have a pure SQL non-relocatable extension that specifies schema in its control file.
Is there anything like https://www.postgresql.org/docs/current/extend-extensions.html#EXTEND-EXTENSIONS-FILES-SCHEMA available?

Looking at the signature of install_extension it is not clear for me.

jim-mlodgenski:
It looks like schema is not exposed in install_extension. Seems to me a logical thing to add

anth0nyleung:
thanks @mkleczek. We are tracking this issue internally.

adamguo0:
Resolved by https://github.com/aws/pg_tle/pull/285

PR #285, opened 2025-01-29
Added schema parameter to install_extension

mkleczek:
Issue #284:

Description of changes:

Added optional _schema_ parameter to install_extension function and implemented adding _schema_ configuration to generated control file.
Updated current extension version to 1.5.0 as the change is not forward compatible. 

### DIFF
--- a/pg_tle--1.4.1--1.5.0.sql
+++ b/pg_tle--1.4.1--1.5.0.sql
@@ -1,0 +1,40 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "CREATE EXTENSION pg_tle" to load this file. \quit
+
+DROP FUNCTION pgtle.install_extension
+(
+  name text,
+  version text,
+  description text,
+  ext text,
+  requires text[]
+);
+CREATE FUNCTION pgtle.install_extension
+(
+  name text,
+  version text,
+  description text,
+  ext text,
+  requires text[] DEFAULT NULL,
+  schema text DEFAULT NULL
+)
+RETURNS boolean
+SET search_path TO 'pgtle'
+AS 'MODULE_PATHNAME', 'pg_tle_install_extension'
+LANGUAGE C;
--- a/pg_tle--1.4.1--1.5.0.sql
+++ b/pg_tle--1.4.1--1.5.0.sql
@@ -25,6 +25,7 @@ DROP FUNCTION pgtle.install_extension
   ext text,
   requires text[]
 );
+
 CREATE FUNCTION pgtle.install_extension
 (
   name text,
@@ -38,3 +39,23 @@ RETURNS boolean
 SET search_path TO 'pgtle'
 AS 'MODULE_PATHNAME', 'pg_tle_install_extension'
 LANGUAGE C;
+
+REVOKE EXECUTE ON FUNCTION pgtle.install_extension
+(
+  name text,
+  version text,
+  description text,
+  ext text,
+  requires text[],
+  schema text
+) FROM PUBLIC;
+
+GRANT EXECUTE ON FUNCTION pgtle.install_extension
+(
+  name text,
+  version text,
+  description text,
+  ext text,
+  requires text[],
+  schema text
+) TO pgtle_admin;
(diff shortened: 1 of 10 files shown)


src/tleextension.c
```diff
@@ -4465,6 +4472,11 @@ pg_tle_install_extension(PG_FUNCTION_ARGS)
 		check_requires_list(reqlist);
 	}
 
+	if (!PG_ARGISNULL(5))
+	{
+		extschema = text_to_cstring(PG_GETARG_TEXT_PP(5));
```
adamguo0, line 4477: I'm getting an uninitialized warning for extschema, maybe we can initialize it to empty string if the argument is not provided.

adamguo0 (commented):
Thank you for your contribution! A couple things:

I think we should return the schema in `pgtle.available_extensions()` and `pgtle.available_extension_versions()`.

We currently override schema to NULL (and relocatable to false) when we actually create the extension (https://github.com/aws/pg_tle/blob/main/src/tleextension.c#L917), so we'll need to remove that line.

Can you rebase against HEAD to get the CI fix?

src/tleextension.c
mkleczek: Added another branch to initialise it to NULL explicitly in case of missing argument. Doing pstrdup here to avoid another conditional below.

mkleczek:
> Thank you for your contribution! A couple things:
> 
> I think we should return the schema in `pgtle.available_extensions()` and `pgtle.available_extension_versions()`.

Looks like schema was already handled in `pgtle.available_extension_versions()`.

Added similar implementation in `pgtle.available_extensions()` but bear with me - I'm not very familiar with Postgres APIs and I am not sure if adding another element to returned tuple requires some more changes somewhere.

> 
> We currently override schema to NULL (and relocatable to false) when we actually create the extension (https://github.com/aws/pg_tle/blob/main/src/tleextension.c#L917), so we'll need to remove that line.
> 

Done.

> Can you rebase against HEAD to get the CI fix?

Done.

adamguo0:
Thanks! I'll take the liberty of running pgindent and adding some unit tests.

adamguo0:
I added unit tests and also reset the change to `pg_tle_available_extensions`. There's some more complexity involved in changing the function return signature so I will do that in a separate PR.

adamguo0:
I encountered some strange behaviour in PG11 where `PG_ARGISNULL` says that the schema parameter is not null, yet `PG_GETARG_DATUM` segfaults because it actually is null. Not sure why there is a discrepancy there but I added a second null check to avoid it.

anth0nyleung:
> I encountered some strange behaviour in PG11 where `PG_ARGISNULL` says that the schema parameter is not null, yet `PG_GETARG_DATUM` segfaults because it actually is null. Not sure why there is a discrepancy there but I added a second null check to avoid it.

Would be interesting to see if we can reproduce this outside of pg_tle, if so maybe we can ask in the upstream community.

anth0nyleung:
> I added unit tests and also reset the change to pg_tle_available_extensions. There's some more complexity involved in changing the function return signature of pgtle.available_extensions() so I will do that in a separate PR.

Do you mind somehow manually verifying that the extension is indeed installed in designated schema? maybe checking pg_proc?

pg_tle--1.4.1--1.5.0.sql
```diff
@@ -0,0 +1,40 @@
...
+
+DROP FUNCTION pgtle.install_extension
+(
+  name text,
+  version text,
+  description text,
+  ext text,
+  requires text[]
+);
+CREATE FUNCTION pgtle.install_extension
```
anth0nyleung, line 28: do we need to re-do this here? https://github.com/aws/pg_tle/blob/22dbe5337936082db1a05e21609417d892667e5f/pg_tle--1.1.1.sql#L273

pg_tle--1.4.1--1.5.0.sql
```diff
@@ -0,0 +1,40 @@
...
+(
+  name text,
+  version text,
+  description text,
+  ext text,
+  requires text[] DEFAULT NULL,
+  schema text DEFAULT NULL
```
anth0nyleung, line 35: we should update our docs for this new parameter support as well

anth0nyleung: actually nvm. i see the regression test has already done it.

anth0nyleung:
> > I added unit tests and also reset the change to pg_tle_available_extensions. There's some more complexity involved in changing the function return signature of pgtle.available_extensions() so I will do that in a separate PR.
> 
> Do you mind somehow manually verifying that the extension is indeed installed in designated schema? maybe checking pg_proc?

actually nvm. i see the regression test has already done it.

pg_tle--1.4.1--1.5.0.sql
adamguo0: Thanks good catch. I'll add a test for this too

test/sql/pg_tle_extension_schema.sql
```diff
@@ -0,0 +1,170 @@
...
+SELECT * FROM pgtle.available_extensions();
+
+CREATE EXTENSION my_tle;
+SELECT my_tle_schema_1.my_tle_func();
+
+-- Cannot drop the schema because the extension depends on it.
+DROP SCHEMA my_tle_schema_1;
+
+-- Clean up.
+DROP SCHEMA my_tle_schema_1 CASCADE;
```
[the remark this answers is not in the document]
adamguo0: The test checks the expected output so if the schema doesn't exist, it should still fail the test

anth0nyleung (approved):

PR #287, opened 2025-02-11
Change return signature of `pgtle.available_extensions()` to match `pgtle.available_extension_versions()`.

adamguo0:
Description of changes: `pgtle.available_extension_versions()` returns more fields than `pgtle.available_extensions()` including the schema field which can now be modified. It doesn't really make sense for the two functions to return different signatures, so add those missing fields to `pgtle.available_extensions()`.

This changes the output returned by `SELECT * FROM pgtle.available_extensions()` so callers should be aware of that. However, the output is identical if the three columns from before are specified, i.e `SELECT name, default_version, comment FROM pgtle.available_extensions()`. This change is part of the 1.5.0 minor version bump.

Add the updated functions to the documentation.

### DIFF
--- a/pg_tle--1.4.1--1.5.0.sql
+++ b/pg_tle--1.4.1--1.5.0.sql
@@ -59,3 +59,25 @@ GRANT EXECUTE ON FUNCTION pgtle.install_extension
   requires text[],
   schema text
 ) TO pgtle_admin;
+
+DROP FUNCTION pgtle.available_extensions
+(
+  OUT name name,
+  OUT default_version text,
+  OUT comment text
+);
+
+CREATE FUNCTION pgtle.available_extensions
+(
+  OUT name name,
+  OUT default_version text,
+  OUT superuser boolean,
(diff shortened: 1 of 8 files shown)


pg_tle--1.4.1--1.5.0.sql
```diff
@@ -59,3 +59,25 @@ GRANT EXECUTE ON FUNCTION pgtle.install_extension
...
+  OUT superuser boolean,
+  OUT trusted boolean,
+  OUT relocatable boolean,
+  OUT schema name,
+  OUT requires name[],
+  OUT comment text
+)
+RETURNS SETOF RECORD
+AS 'MODULE_PATHNAME', 'pg_tle_available_extensions'
+LANGUAGE C STABLE STRICT;
```
adamguo0, line 83: No need to revoke permissions here since this is callable by unprivileged users

sibowu-aws (approved):

anth0nyleung (approved):

PR #295, opened 2025-03-19
Remove idx_offset optimization from clientauth.

adamguo0:
Description of changes:

There was a math bug with this optimization that caused incoming connections to occassionally be missed by a worker.  This optimization doesn't seem necessary and makes the code harder to understand and maintain, so just get rid of it.

This is the root cause of some recent CI failures, e.g. https://github.com/aws/pg_tle/actions/runs/13939294913

Details:

- There are 256 slots in the clientauth request queue. If there are fewer than 256 clientauth workers (let's say `n` workers), we assign every `n` slots to each worker.
    - e.g. if there are 20 workers (`n = 20`), then worker `i` will be assigned slots `i`, `i + 20`, `i + 40`, and so on.
- Each clientauth worker uses an [`idx_offset`](https://github.com/aws/pg_tle/blob/main/src/clientauth.c#L417-L421) to try and randomize the sequence in which it will check its assigned slots. This is an optimization to improve the distribution of backend processes that are picked up by each worker during connection storms.
- There is a bug with the logic. When `idx_offset` is non-zero and the incoming request is less than the number of clientauth workers, the worker will never pick that connection.
    - The reason is that [clientauth.c:424](https://github.com/aws/pg_tle/blob/main/src/clientauth.c#L424) was meant to reset idx to the first slot assigned to this worker, but it doesn't do that if 256 is not divisible by clientauth_num_workers. I think we primarily tested with divisible numbers (like 1 or 2 or 64) in the past.
    - The failing test tests with 60, which 256 is not divisible by.
    - The test only fails when the incoming connection happens to be assigned to one of the first 60 slots.
    - In theory we should have seen this bug before, but the most recent PR added more tests after the test that sets num workers to 60, so we only started seeing it now.
- IMO this optimization is not necessary and makes the code harder to read and maintain. We don't have data suggesting that it was helpful anyway. I'm proposing to just get rid of the idx_offset logic and reduce the potential for other weird calculation issues.

### DIFF
--- a/src/clientauth.c
+++ b/src/clientauth.c
@@ -369,9 +369,6 @@ clientauth_launcher_main(Datum arg)
 {
 	int			bgw_idx = DatumGetInt32(arg);
 
-	/* Keeps track of which of its entries the worker will look at first. */
-	int			idx_offset = 0;
-
 	/* Establish signal handlers before unblocking signals */
 	pqsignal(SIGHUP, clientauth_sighup);
 	pqsignal(SIGTERM, die);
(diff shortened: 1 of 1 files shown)


esiaero (approved):
LGTM

adamguo0:
Pushed a commit to account for upstream test change https://github.com/postgres/postgres/commit/4f7f7b0

PR #308, opened 2026-06-11
Validate identifiers substituted into extension scripts

jim-mlodgenski:
Issue #, if available:

Description of changes:

`execute_extension_script()` substitutes `@extowner@` and `@extschema@`
into the extension script, filtering each value through
`quote_identifier()`. Certain characters (`"` `$` `'` `\`) cannot be
quoted consistently both inside and outside of string literals, so a
name containing one of them can produce a script that does not parse the
way the extension author intended once the substitution is performed.

This change rejects substitution when the owner or schema name contains
any of these characters instead of producing such a script, matching the
identifier handling that PostgreSQL core performs for the same
substitutions. The `pg_tle_injection` regression test is extended with
coverage for both the owner and schema cases, including a check that
ordinary names continue to work.

### DIFF
--- a/test/sql/pg_tle_injection.sql
+++ b/test/sql/pg_tle_injection.sql
@@ -175,6 +175,57 @@ $_pg_tle_$
 $_pg_tle_$
 );
 
+-- @extschema@ and @extowner@ substitutions are filtered through
+-- quote_identifier(). A schema or owner name containing a character that
+-- cannot be consistently quoted inside and outside of string literals (any of
+-- " $ ' \) must be rejected rather than substituted into the script.
+
+-- An extension whose script references @extschema@ cannot be created into a
+-- schema whose name contains a quoting-relevant character.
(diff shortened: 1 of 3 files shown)


test/sql/pg_tle_injection.sql
```diff
@@ -175,6 +175,57 @@ $_pg_tle_$
 $_pg_tle_$
 );
 
+-- @extschema@ and @extowner@ substitutions are filtered through
+-- quote_identifier(). A schema or owner name containing a character that
+-- cannot be consistently quoted inside and outside of string literals (any of
+-- " $ ' \) must be rejected rather than substituted into the script.
```
sibowu-aws, line 181: nit: guess it is still beneficial to test $ and \ ?

anth0nyleung:
LGTM. Could we fix the CI for master branch with pgindent?

Issue #301, opened 2025-07-25
How to set required schema when upgrading from 1.4 to 1.5


jnasbyupgrade:
Is there any way to update an existing extension to mark it as requiring a specific schema when upgrading from 1.4 to 1.5?

For the use case I have in mind there are configuration tables, so simply dropping and re-creating the extension would be problematic.

jim-mlodgenski:
You can likely use the native PG functions for config tables.
https://www.postgresql.org/docs/current/extend-extensions.html#EXTEND-EXTENSIONS-CONFIG-TABLES

jnasbyupgrade:
My question isn't how to mark a table as config; I mentioned that I have config tables because it means simply dropping and re-creating the extension is problematic.

The real question is: if I create an extension in TLE 1.4 (which doesn't support specifying a schema), how can I update that extension in a newer version of TLE (that includes schema support)? It appears there's no way to do that (unless I'm missing something)... would there be interest in adding that ability?

PR #311, opened 2026-06-23
Add pgtle.set_extension_schema() to set an extension's schema

jim-mlodgenski:
Issue #, if available: #301

Description of changes:

Schema support added in pg_tle 1.5 stores the target schema in an
extension's control function, but there was no way to set or change it
afterward. An extension installed before 1.5, or with the wrong schema,
could only adopt one by uninstalling and reinstalling, which drops any
data in its tables.

This change adds pgtle.set_extension_schema(name, schema), which
rewrites just the control function in place. It takes the schema as
text (matching install_extension's schema argument, which likewise only
records the name) and accepts NULL to clear it; an empty string is
rejected. The change affects future CREATE EXTENSION calls; an
already-created extension is left untouched. The new function ships in a
1.5.2--1.5.3 upgrade script, with regression coverage in
pg_tle_extension_schema and documentation in 03_managing_extensions.md.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 EXTENSION = pg_tle
-EXTVERSION = 1.5.2
+EXTVERSION = 1.5.3
 
 SCHEMA = pgtle
 MODULE_big = $(EXTENSION)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -9,7 +9,8 @@ OBJS = src/tleextension.o src/guc-file.o src/feature.o src/passcheck.o src/uni_a
 EXTRA_CLEAN	= src/guc-file.c pg_tle.control pg_tle--$(EXTVERSION).sql
 DATA = pg_tle.control pg_tle--1.0.0.sql pg_tle--1.0.0--1.0.1.sql pg_tle--1.0.1--1.0.4.sql pg_tle--1.0.4.sql pg_tle--1.0.4--1.1.1.sql \
 			 pg_tle--1.1.0--1.1.1.sql pg_tle--1.1.1.sql pg_tle--1.1.1--1.2.0.sql pg_tle--1.2.0--1.3.0.sql pg_tle--1.3.0--1.3.3.sql \
-			 pg_tle--1.3.3--1.3.4.sql pg_tle--1.3.4--1.4.0.sql pg_tle--1.4.0--1.5.0.sql pg_tle--1.5.0--1.5.2.sql
+			 pg_tle--1.3.3--1.3.4.sql pg_tle--1.3.4--1.4.0.sql pg_tle--1.4.0--1.5.0.sql pg_tle--1.5.0--1.5.2.sql \
```
jnasbyupgrade, line 12: FWIW, the list of update files could be replaced with `$(wildcard *--*--*.sql)` (pgxntool does [pretty much that](https://github.com/Postgres-Extensions/pgxntool/blob/2.0.2/base.mk#L40); modulo assuming all .sql files live in `sql/`).

jnasbyupgrade (approved):
Thanks for handling this!

PR #317, opened 2026-07-22
Fix login hang when pgtle.clientauth_db_name names a nonexistent database

zjyzh:
### Bug
`pgtle.clientauth_db_name` is `PGC_POSTMASTER`. If it names a nonexistent (or later-dropped) database, every clientauth worker FATALs and every login hangs on a rendezvous no worker completes. Nobody can log in to fix it.

### Root cause
`clientauth_hook` enqueues and waits on `client_cv` with no timeout and no check that the shard has a live worker. It can't look up `pg_database` itself — it runs before `InitPostgres`.

### Fix
Give the hook a way to know whether a worker will ever answer, so it stops waiting forever.

Each background worker now flips a shared "I'm alive" flag once it has successfully attached to the database, and flips it back on exit. When a login arrives, the hook first checks that flag:

- If no worker comes up within 3 seconds, the hook stops waiting and falls back to the existing behavior — `enable_clientauth = 'on'` accepts the connection, `'require'` rejects it with a clear message saying to check `clientauth_db_name`.
- If the worker was alive but dies while handling the request, the hook notices within about a second and takes the same fallback.

The flag is per-worker (not a single "any worker up" bit) because requests are split across workers by client PID, so the hook has to check the specific worker its request will route to.

## Branches
`aws/pg_tle:main` — `fix/clientauth-missing-dbname`, commit `f7588c2`.

## Testing
Three new TAP subtests in `004_pg_tle_clientauth.pl` (#20/#21/#22), all bounded `timeout => 15`.

- #20: `on` + missing DB → accepts.
- #21: `require` + missing DB → clean FATAL.
- #22: `DROP DATABASE ... WITH (FORCE)` → new logins don't hang.

Patched: 6/6 pass in ~20s. Unpatched baseline: 77s hang, bails.

## License

### DIFF
--- a/include/compatibility.h
+++ b/include/compatibility.h
@@ -615,6 +615,37 @@ CastCreate(Oid sourcetypeid, Oid targettypeid, Oid funcid, char castcontext,
 #endif
 
 /*
+ * PostgreSQL 13 introduced ConditionVariableTimedSleep(). On PG 12 we
+ * emulate it with WaitLatch(); the wait-loop protocol (PrepareToSleep +
+ * CancelSleep) is otherwise identical, and returns true iff the timeout
+ * elapsed.
+ */
+#include "storage/condition_variable.h"
+#include "storage/ipc.h"
+#include "storage/latch.h"
+#include "miscadmin.h"
+#if PG_VERSION_NUM < 130000
+static inline bool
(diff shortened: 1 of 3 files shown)


jim-mlodgenski:
This is broken on PG12. Since there is still LTS support around for 12, we should resolve that before merging

PR #318, opened 2026-07-23
Emit WARNING instead of NOTICE in extension scripts

zjyzh:
### Bug
Seven `RAISE NOTICE` calls in the pg_tle install scripts never reach the client or the log. Operators get no guidance when they hit them.

### Root cause
Postgres's `execute_extension_script()` clamps `client_min_messages` and `log_min_messages` to at least `WARNING` for the duration of `CREATE/ALTER EXTENSION`. NOTICE-level messages raised from inside an extension script are dropped.

### Fix
Promote the seven NOTICE calls to WARNING so the guidance survives the clamp. Sites: the "role already exists" guard in `pg_tle--1.0.0.sql` / `1.0.4.sql` / `1.1.1.sql`, and four config-guidance messages inside `pgtle.register_feature` in `pg_tle--1.3.4--1.4.0.sql`.

## Branches
`aws/pg_tle:main` — `raise-warning-in-extension-scripts`, commit `c248746`.

## Testing
`make installcheck` on PostgreSQL 17.10, cold cluster: 10/10 regression + 107/107 TAP pass. Seven `.out` files updated to match the new severity label.

## License

### DIFF
--- a/pg_tle--1.0.0.sql
+++ b/pg_tle--1.0.0.sql
@@ -278,7 +278,7 @@ BEGIN
       SELECT FROM pg_catalog.pg_roles
       WHERE  rolname = 'pgtle_admin') THEN
 
-      RAISE NOTICE 'Role "pgtle_admin" already exists. Skipping.';
+      RAISE WARNING 'Role "pgtle_admin" already exists. Skipping.';
    ELSE
       CREATE ROLE pgtle_admin NOLOGIN;
    END IF;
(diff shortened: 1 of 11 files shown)


anth0nyleung (approved):

jkatz:
Why not put `SET client_min_messages = 'NOTICE'` in the extension script? These aren't necessarily `WARNING` level issues (e.g. the `pgtle_admin` user already existing).

anth0nyleung:
@jkatz raised a good point, please also investigate the CI/CD.

